Issue #5, opened 2016-01-31
Multicolumn has a standard width, single column a variable width. 


ludobrands:
```
<mj-body> 
  <mj-section>
    <mj-column >
      <mj-divider border-color="#F45E43"></mj-divider>
      <mj-text font-size="20px" color="#F45E43" font-family="helvetica">Lorem ipsum dolor sit amet, consectetur adipiscing elit. Vestibulum nibh lorem, fermentum sit amet leo sed, consectetur egestas ligula. </mj-text>
    </mj-column>
  </mj-section>
  <mj-section>
    <mj-column >
      <mj-divider border-color="#F45E43"></mj-divider>
      <mj-text font-size="20px" color="#F45E43" font-family="helvetica">Hoj</mj-text>
    </mj-column>
  </mj-section>
</mj-body>
```

vs

```
<mj-body>
  <mj-section>
    <mj-column >
      <mj-divider border-color="#F45E43"></mj-divider>
      <mj-text font-size="20px" color="#F45E43" font-family="helvetica">Lorem ipsum dolor sit amet, consectetur adipiscing elit. Vestibulum nibh lorem, fermentum sit amet leo sed, consectetur egestas ligula. </mj-text>
    </mj-column>
    <mj-column >
      <mj-divider border-color="#F45E43"></mj-divider>
      <mj-text font-size="20px" color="#F45E43" font-family="helvetica">Hoj</mj-text>
    </mj-column>
  </mj-section>
  <mj-section>
    <mj-column >
      <mj-divider border-color="#F45E43"></mj-divider>
      <mj-text font-size="20px" color="#F45E43" font-family="helvetica">Hoj</mj-text>
    </mj-column>
    <mj-column >
      <mj-divider border-color="#F45E43"></mj-divider>
      <mj-text font-size="20px" color="#F45E43" font-family="helvetica">Lorem ipsum dolor sit amet, consectetur adipiscing elit. Vestibulum nibh lorem, fermentum sit amet leo sed, consectetur egestas ligula.</mj-text>
    </mj-column>
  </mj-section>
</mj-body>
```

Removing the  redundant `<mj-column >  </mj-column >` in the first example results in aligned text  but there is no reason why a single column should behave different from multi column.  
Adding width=100% to the column has the desired effect. This should be the default for a single column.

iRyusa:
Hello !

I think this is because it's missing a media query, I added one for single column layout. 

Layout from exemple looks like this : 
[...]

PR #6, opened 2016-02-01
Missing a default media query when no width provided

iRyusa:
When using `mj-column` for a single column layout with no `width` it doesn't behave as a `width="100%"`as expected. Poke #5


iRyusa:
Waiting for #5 to merge it

iRyusa:
Closing the PR because a better workaround has been found

PR #8, opened 2016-02-02
Width parser, decorators, background for outlook button

iRyusa:
It will be easier to manage width units for column and image.

Outlook supports only `background` instead of `background-color`

Move decorators to `/components/decorators`


robink:
:+1:

Issue #44, opened 2016-02-11
Missing padding attribute on Section element


Gilbert09:
The section element has `padding-left`, `padding-right` etc but no `padding` which makes it slightly annoying having to set the padding on all 4 sides.

iRyusa:
Now added in `1.2.0`, documentation should be updated today

PR #47, opened 2016-02-11
Missing padding attribute on Section element

Gilbert09:
#44

### DIFF
--- a/src/components/Section.js
+++ b/src/components/Section.js
@@ -56,7 +56,8 @@ class Section extends Component {
         paddingTop: mjAttribute('padding-top'),
         paddingBottom: mjAttribute('padding-bottom'),
         paddingRight: mjAttribute('padding-right'),
-        paddingLeft: mjAttribute('padding-left')
+        paddingLeft: mjAttribute('padding-left'),
+		padding: mjAttribute('padding')
       }
     })
   }
--- a/src/components/Section.js
+++ b/src/components/Section.js
@@ -53,11 +53,11 @@ class Section extends Component {
       td: {
         fontSize: 0,
         verticalAlign: mjAttribute('vertical-align'),
+		padding: mjAttribute('padding'),
         paddingTop: mjAttribute('padding-top'),
         paddingBottom: mjAttribute('padding-bottom'),
         paddingRight: mjAttribute('padding-right'),
-        paddingLeft: mjAttribute('padding-left'),
-        padding: mjAttribute('padding')
+        paddingLeft: mjAttribute('padding-left')
       }
     })
   }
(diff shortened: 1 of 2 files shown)


src/components/Section.js
```diff
@@ -56,7 +56,8 @@ class Section extends Component {
         paddingTop: mjAttribute('padding-top'),
         paddingBottom: mjAttribute('padding-bottom'),
         paddingRight: mjAttribute('padding-right'),
-        paddingLeft: mjAttribute('padding-left')
+        paddingLeft: mjAttribute('padding-left'),
+        padding: mjAttribute('padding')
```
iRyusa: This should be on top of every paddingLeft/Right/Top/Bottom. They must override `padding` if they're present like so :
```xml
<mj-column padding="20 0" padding-left="40px">
```

iRyusa:
Hi thanks for contributing @Gilbert09 !

We'll have to do this on `column` and `column-element` too. 
But this need some work to be supported for `Image` !

src/components/Section.js
Gilbert09: @iRyusa Should this also be the same on `MJMLColumnElement`? As currently it's at the bottom of the stack too.

iRyusa: I haven't seen that `MJMLColumnElement` has already `padding` thanks for noticing ! Yes it would be nice to !

Gilbert09: Just moved them both to the top of the padding attribute stack :smile:

src/components/Section.js
```diff
@@ -53,6 +53,7 @@ class Section extends Component {
       td: {
         fontSize: 0,
         verticalAlign: mjAttribute('vertical-align'),
+		padding: mjAttribute('padding'),
```
iRyusa: Good ! Can you just fix your editor to put space instead of tabs so this can be merged ? Thanks !

Gilbert09: Already done :smile:

iRyusa:
I would like to merge it, but I think this is a start of rework how mjml handles padding. 

CSS padding will be overriden by any **default** padding-left/right/etc, and we don't want to. But we do want to override `padding` when it comes from markup input like so : `<mj-text padding="15" padding-left="20" />. 

If you have any idea on how to implement that !

I think we need to open a new `padding-rework` branch and merge your work in it are you ok with this ? @Gilbert09

Gilbert09:
@iRyusa Yep, I'm perfectly happy with that. I've been trying to think how I can also extend Mjml to allow some kind of class support with SASS. Not really come up with anything yet though.

It would be interesting to hear what you plans are for the near future of Mjml, as I'm already making emails using Mjml (with the help of gulp) for production in my company.

Gilbert09:
@iRyusa So I made the fatal mistake of doing this pull request from my master branch - so I'm gonna close this one and reopen another.

PR #91, opened 2016-02-24
Stream support for cli

zakdoek:
Added stream support for cli.

Functions as before but with 2 extra option flags:
- `-s, --stdout` to output to stdout instead of file
- `-i, --stdin` to get the mjml source from input stream


GuillaumeBadi:
Thanks @zakdoek !
That's awesome!

PR #93, opened 2016-02-24
Correct text color when parent has color too

meriadec:
see https://mjml.io/try-it-live/31f19dee07b3a16ac1b0

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -129,7 +129,7 @@ class ${name} extends Component {
    * Build your styling here
    */
   getStyles() {
-    const { mjAttribute, color } = this.props
+    const { mjAttribute, parentColor } = this.props
 
     return _.merge({}, this.constructor.baseStyles, {
       text: {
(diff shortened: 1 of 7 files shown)


meriadec:
tl;dr here is the case:

```
mj-body (color: red)
  mj-text (color: blue)    ==> should be blue
  mj-text                  ==> should be red
```

but actually, we can't check if color has been set or not, because [color is always set in default attributes](https://github.com/meriadec/mjml/blob/579ee54b52180776b730827da5b46ef475c19d01/src/components/Text.js#L14). So in which case should component take color from parent / from its attributes?

iRyusa:
So maybe we need to change how `mjAttribute` works, now default attribute should stay at `state.elem.defaultAttribute` and 

```
mjAttribute(name) {
      return_.filter([this.props[name], this.state.getIn(['elem', 'attributes']).get(name), this.state.getIn(['elem', 'defaultAttributes']).get(name) ], function(e){ return e != undefined })[0] 
}
```

meriadec:
If you are ok with this solution, I'll update `Button`, `Social` and `List` to have `color` in `defaultAttributes` instead of `attributes`

meriadec:
closing this, not accurate anymore with current mjml engine

PR #105, opened 2016-02-29
Remove 'color' props from childDefaultProps

swibge:
Remove color props from childDefaultProps 'cause if a parent component need to "renderChildren", default child color attribute is not available


iRyusa:
Goes with @meriadec PR, I think we can safely delete it... :+1: then sorry for your PR @meriadec :<

Issue #86, opened 2016-02-23
max-width property really useful for mj-image?


epayet:
mj-image sets by default the css property `max-width` : 

```javascript
getStyles() {
    const { mjAttribute } = this.props

    return _.merge({}, this.constructor.baseStyles, {
      img: {
        border: mjAttribute('border'),
// Here
        maxWidth: this.getContentWidth(),
        height: mjAttribute('height')
      }
    })
  }
```

I have a case where it is much better looking on mobile without this property.

Here is the mobile version, the first image is the MjImage component without max-width, and the second one is the default MjImage component:

![mobile](https://gist.githubusercontent.com/epayet/fa058ac45dd748e5c6e0/raw/1e5d5694857cc2ee7cc45b9772d2b18ab6491df7/mj-image-max-width-mobile2.png)

And here is the desktop version unaffected:

![desktop](https://gist.githubusercontent.com/epayet/fa058ac45dd748e5c6e0/raw/b157c81f27e72ab3e479af5266eba860aaa97a90/mj-image-max-width-desktop.png)

So in my case, I prefer the version without max-width. There is too much white space on the default one. Is this an expected behaviour? Should we have this optional? What are your thoughts on that? thanks

iRyusa:
Hi @epayet,

Did you try to set `padding="0"` on your `mj-image` ? Because `max-width` is computed as such: 
`min(columnWidth/widthAttr) - paddingLeft - paddingRight)` in order to avoid a column overflow.

epayet:
I tried, in the example above, I am using `padding="10px"`, changing it to `padding="0"` just gave me 20px more, but still does not look like I want it to (the first tile) :disappointed: 

Here is what I changed to have the first tile on MjImage:

```javascript
getStyles() {
    const { mjAttribute } = this.props

    // Removed max-width property
    return _.merge({}, this.constructor.baseStyles, {
      img: {
        border: mjAttribute('border'),
        // removed this
        //maxWidth: this.getContentWidth(),
        height: mjAttribute('height')
      }
    })
  }
```

and

```javascript
renderImage() {
    const { mjAttribute } = this.props

    // Changed here
    const img = (
      <img
        alt={mjAttribute('alt')}
        border="0"
        src={mjAttribute('src')}
        style={this.styles.img}
        // instead of width={this.styles.img.maxWidth}
        width={this.getContentWidth()}
        height={this.styles.img.height} />
    )
...
```

iRyusa:
[...]

PR #145, opened 2016-03-22
Fix and remove mj-image default css width

Tlapi:
There is no reason to add 100% default css width

### DIFF
--- a/src/components/Image.js
+++ b/src/components/Image.js
@@ -29,8 +29,7 @@ class Image extends Component {
       border: 'none',
       display: 'block',
       outline: 'none',
-      textDecoration: 'none',
-      width: '100%'
+      textDecoration: 'none'
     }
   }
 


iRyusa:
Hi @Tlapi !

I'd prefer to keep the image 100% to avoid the overflow. I'd rather add a `width` on `td` to keep a safety for big images. Does it make sense to you ?

Tlapi:
Hey @iRyusa,

if you want to avoid overflow, set max-width. But now, when you set width of image to something less then 100%, it gets ignored, as css inline style is prefered over width attribute.

iRyusa:
`Max-width` is a pain see #86 :/

`Width` attribute is only here for Outlook that doesn't support `style width` for `img`
Adding the width on the `td` should solve your issue since the 100% will match the container

Tlapi:
Ok, i have checked and tested, seems fine to set `width` attribute on `td`. I will close this PR and open a new one.

Tlapi:
Uhm, this gets a bit messy. So i will leave that up to you to keep your standart.

janus-reith:
No it is not fine.
Before i could simply set the width of my image to 560px, achieving 20px padding at each side, while still having the full width on mobile view.

Setting the column width to 560px breaks the mobile view as it will not scale anymore.
Adding 20px padding to the image will also make it way smaller in mobile view.

Adding the ability of mobile/desktop only styles could be a better long term solution...
Using mj-raw everytime I need the is a bit messy

iRyusa:
> Adding the ability of mobile/desktop only styles could be a better long term solution... 
Fact is, no so much email clients supports media queries. A PR will be open shortly to fix the width issue

PR #154, opened 2016-03-24
Table layout property to set it as 'fixed' e.g.

slantz:


iRyusa:
Looks good to me :+1: !

lohek:
:+1:

faergeek:
I'm sorry for probably stupid question (I'm new to mjml), but how do I set `table-layout` on `mj-column`?

I tried `<mj-column table-layout="fixed">`. No `table-layout: fixed` on rendered tag :-(

ngarnier:
hey @faergeek, `table-layout` is an attribute available on `mj-table`: https://mjml.io/documentation/#mjml-table

faergeek:
@ngarnier Hi, thanks for the answer! What can I do to have the same for `mj-column`? Right now I've got issues with Thunderbird which are solved if I set `table-layout: fixed` on `mj-column` (checked after connecting Firefox's remote debugger to Thunderbird).

ngarnier:
hey @faergeek, sorry for the late answer, missed your response. So using `table-layout:fixed` solves the issue on Thunderbird without leading to other issues?

How are you setting it now, straight into the HTML?

faergeek:
@ngarnier I set it in firefox devtools connected to Thunderbird. I did not work around it in other way because that client is not in priority. But this fix is pretty simple so would be great if I could do it.

ngarnier:
Alright, well if you want to see how it behaves you can try to fork MJML and edit [mj-column](https://github.com/mjmlio/mjml/blob/master/packages/mjml-column/src/index.js) to set `table-layout` to `fixed` like it was done on this PR. Two options here:

- Set it as a default style in the output HTML
- Set it as an attribute

Then you can see if it has other unexpected impacts that could break responsiveness elsewhere.

Does it help?

faergeek:
Yeah, I'll try it a bit later. Thanks for help 👍

PR #158, opened 2016-03-25
Mj socail div alignment

slantz:


iRyusa:
Hi @slantz 

We had an internal discussion over `align` attribute, it will now be applied as a `text-align` (almost as your previous PR, sorry about that) on the `td` of a column element. 

It will be ok on the next release (2.0 for now I guess) ! 

Closing it, sorry for the delay

PR #161, opened 2016-03-26
Better productivity with SpeedyTag

martindebrunne:
Now we can use a fast similar slim tag. I named it SpeedyTag and i used .mjml.fit extension to distinct .mjml file. But that doesn't matter. You just need to use SpeedyTag format in a file. 
It's based on mjml tags. 
See files examples in test/assets.  You can use same commands lines to render a html file and give the right file.

### DIFF
--- a/src/documentParser.js
+++ b/src/documentParser.js
@@ -2,6 +2,7 @@ import _ from 'lodash'
 import { endingTags } from './MJMLElementsCollection'
 import { ParseError, EmptyMJMLError, NullElementError } from './Error'
 import dom from './helpers/dom'
+import { speedyTagToMJM, presenceOfSpeedyTag } from './helpers/speedyTag'
 
 /**
  * Avoid htmlparser to parse ending tags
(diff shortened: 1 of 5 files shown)


martindebrunne:
Should I change name extension in order to pass tests ?

iRyusa:
Wow ! Looks really impressive :+1:

ngarnier:
Hey @martindebrunne, 

Thanks for the PR, it's really interesting. Unfortunately, we can't merge it because it would mean a heavy dependency to `mjml-core`. 

A better approach would be to make it a plugin / tool around MJML, maybe with integrations in text editors such as Atom/Sublime Text? What do you think?

martindebrunne:
Thank you for your reply. I think a tool around MJML can be nice and some plugins for text editors in order to 'check syntax and run the MJML tool' can be fine for the development and make easy the integration in a lot of text editors. 
When I will have time I will try to do it.

ngarnier:
Hi, I'm closing this PR as it's based on a version of MJML which is very outdated. 

As discussed at the time, we've worked with the help of the community to reach a better productivity thanks to plugins and tools. Here are a few of the ones available:
- Atom packages: https://atom.io/users/mjmlio
- Sublime package: https://packagecontrol.io/packages/MJML-syntax
- Collection of tools (Vim plugin, Gulp, Grunt...): https://mjml.io/community

Thanks again!

PR #165, opened 2016-03-29
Fix Outlook VML Background.

janus-reith:
Changed <v:fill type from type="tile" to type ="frame".
A 600px background will now be displayed correct on a 600px section.

The issue:
When using a background image smaller than the size of the section, it will be stretched to fit when using type="frame".

-> Need to utilize background-size or background-repeat to decide if frame ot tile is used.
I'd appreciate your advice here on which to use and how to keep thing compatible.

**Please dont merge this before we can make sure that it wont break backgrounds for many others.**

### DIFF
--- a/src/mjml2html.js
+++ b/src/mjml2html.js
@@ -82,7 +82,7 @@ const fixOutlookLayout = $ => {
 
     $(this).before(`<!--[if gte mso 9]>
       <v:rect xmlns:v="urn:schemas-microsoft-com:vml" fill="true" stroke="false" style="width:${width}px;">
-        <v:fill origin="0.5, 0" position="0.5,0" type="tile" src="${url}" />
+        <v:fill origin="0.5, 0" position="0.5,0" type="frame" src="${url}" />
         <v:textbox style="mso-fit-shape-to-text:true" inset="0,0,0,0">
       <![endif]-->`)
 


janus-reith:
Something seems wrong there, I hardly believe that this little commit made the build fail on njs 4.2?

janus-reith:
Okay, when using a background image smaller than the size of the section, it will be stretched to fit when using type="frame".

-> Need to utilize background-size or background-repeat for that.

iRyusa:
Looks good to me :+1: 
Can we add some before/after screenshot to this PR ?

janus-reith:
Sorry for the late reply, here is a simple example using the standard code that is shown when opening the try-it-live editor and adding this 600px background image:
colormeter


What it should look like(and does in most clients):
right


What it looks like in Outlook:
outllook

What it looks like in Outlook after replacing every type="tile" with type="frame" in the output:
outllook_fixed

**The issue while this can not be merged easily:** When using type="frame", images that have a width of less than 600px will not be tiled, but stretched to fit instead. But with the current type="tile", images that have the exact size do not fit at all, like seen above.

Also, note how the logo and the divider are still shifted to the left in Outlook,
because I used the background-image property.
Although these screenshots are from the live editor, I just tested and had the same results in 1.3.3 cli.

iRyusa:
Hello @janus-reith ,

Can you update this PR, I would like to merge it on MJML 2.0  👍

m2ses:
Hi @janus-reith 
I can't get my images to work correctly in Outlook.com even when changing the Fill to frame, my large images get clipped and my small images repeat even with background-repeat: no-repeat. Any ideas?. Thanks in advance for your help with this.

iRyusa:
Hi @m2ses 

It's a separate issue, and VML is only used on desktop outlook, not outlook.com. 
Note that every background property ( size / repeat ) aren't supported on outlook.com

m2ses:
hi iRyusa, thanks for the response. 
The VML code appears to be working fine in outlook.com, as the desired text is being added correctly to the cell and over the image without any problems, the problem is the image is not automatically resized to the size of the rectangle which I though would happened when using v:fill type"frame" as indicated by emailsonacid and others. Is there a way to achieve automatically resizing the image that you know of, may with imageData or another way? 
Thank you in advance for your response.

m2ses:
here is a test example
test

iRyusa:
As I said, outlook.com **totally ignores** the VML because it's only for Outlook **Desktop** ( 2003/2007/2011/2016...)
And as I said Outlook.com doesn't support at all any css background attribute so there's no way to do that. You can read more about CSS attributes support for client on that list https://www.campaignmonitor.com/css/ but last update was in 2014.

bago:
Outlook.com should support BACKGROUND html attribute on TABLE/TD: IIRC this is the only supported background there. I don't remember if Outlook.com mobile app is consistent with the webmail or not.

iRyusa:
@bago Yes, that what we use to enable background on outlook https://github.com/mjmlio/mjml/blob/master/packages/mjml-section/src/index.js#L173

iRyusa:
Well this haven't received any update for awhile now, and we're not sure how to allow both behavior. We don't want to add an extra desktop outlook attribute only and the default type match more current usage in MJML.

However if someone has a neat idea on how to allow both `type` in a clean way, then feel free to reopen on MJML4 branch `next` !

Issue #157, opened 2016-03-24
Rendering Components without a Body


arist0tl3:
Hey guys, loving this so far!

Out of curiosity, is there a way to render individual components, without a body?

For instance, right now for a component I can do something like:

    <mj-body>
      <mj-section>
        <mj-text>
          Hi!
        </mj-text>
      </mj-section>
    </mj-body>

and then grab everything in between the `<div>` tags inside the `<body>`, save that to a variable, and repeat. Finally, I do a render of a block of mjml with _all_ of my possible components, so that I have a header that addresses all of the columns, etc. and wrap that around all of my other content. Which works. Which is great! But from an efficiency standpoint, not so much. What I'd like to be able to render is from:

    <mj-section>
      <mj-text>
      </mj-text>
    <mj-section>

so I'm not stripping out the body/header on each render.

It seems like there are references/dependencies on the parent elements that may prevent this right now? Which is why the `Null element found in mjmlElementParser` error is triggered?

Thanks in advance for you thoughts here!

jquinter:
in fact, there is dependencies: if you check the generated html, mainly css classes and stuff.

I think this proposal may cover what you asked: #78 , so please check it and leave a thumb up if it would work for you

arist0tl3:
@jquinter Thanks for the response, I thumbed that thread -- looks interesting!

arnaudbreton:
Hi @arist0tl3, thanks for sharing. Thanks @jquinter for helping here!

The body, acting as a root for the document is indeed required and we won't be able to bypass it. Also, in order to provide you with a responsive HTML code, we need the full HTML generated. 

Since each component is actually a ReactJS one, I believe there is a way to render it individually through React directly, allowing you just to grab the final HTML, and mix everything later on including the full body we generate.

May I ask what is your exact use-case here, in order to see if there is not another solution, already implemented to help you here? 

Thanks!

arist0tl3:
@arnaudbreton Thanks for the response!

I agree that my use is likely a bit of an edge case and probably wouldn't justify putting in a considerable effort. That being said, though:

A current project is building out a way for users to create newsletters, transactional emails, etc. Right now, my workflow looks like:

1) Designer comes up with a new theme/template
2) I take each of our components and render them individually (mjml->html)
3) I put all of the components together and render them to get the header (mjml->html)
4) I place the placeholder variables for the content in the html-rendered components and kind of piece things together.

[...]

Issue #153, opened 2016-03-24
<!-- ... --> HTML Comments break the code inside an Ending Tag


janus-reith:
I used to comment my mjml code with  tags.
Since 1.3 it seems like they get parsed somehow,
in some parts they appear as text, in some they just break the funcionality of the section after the comment.

I tried CSS comments, they also don't work.
If HTML Comments was just a thing that worked unintentionally,
you can see this as a feature request.

iRyusa:
Hi @janus-reith 

Looks like to work for me on `1.3.3` can you give me your MJML ?

janus-reith:
This is the code that  worked before the update.
(In this code you can also see how the image width is ignored.)

[code.txt](https://github.com/mjmlio/mjml/files/187871/code.txt)

iRyusa:
Updating the issue title, this might be an issue with htmlparser

mattpirelli:
I second this issue. I've updated just yesterday to 1.3.3 and can confirm that HTML comments inside `<mj-text>` break the rendering. Well, it [fails spectacularly](http://i.imgur.com/seWzWLm.jpg) :tada: 



Strangely enough, HTML comments inside other components (namely, `<mj-button>`) don't seem to interfere with the rendering.

PR #168, opened 2016-03-29
Bad decodeEnties, bad.

lohek:
Wrong issue related to #153 not #157 ;_;


lohek:
It will be directly in the next major release

PR #186, opened 2016-04-08
Fix small font-sizes 

keroumi:
When rendered you see font-size:0px; or 1px everywhere, that increases spam score that may affect mail delivery.

Thanks!


lohek:
Hi !

Maybe you are wrong, please re-create your issue in issues section https://github.com/mjmlio/mjml/issues.

Thanks.

PR #192, opened 2016-04-17
Added default values for invoice price and quantity

kachar:
When user do not provide valid values for `mj-invoice-item` object mjml results in error.

Covered cases
```
<mj-invoice-item name="Some name" price="123" quantity="" />
<mj-invoice-item name="Some name" price="123" quantity="not-int" />
```

https://mjml.io/try-it-live/dac50d64d46b4c376dba3f449483faa0

### DIFF
--- a/src/components/Invoice/index.js
+++ b/src/components/Invoice/index.js
@@ -102,8 +102,8 @@ class Invoice extends Component {
     const currency = this.currency
 
     const total = this.items.reduce((prev, item) => {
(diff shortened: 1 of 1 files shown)


iRyusa:
Hello @kachar ! 

Can you update your PR to reflect this change on MJML 2.0 ? Thanks a lot !

kachar:
I've created a new PR #211 against the new structure of MJML 2.0

Issue #160, opened 2016-03-25
Gutter between columns / body


EmmanuelBeziat:
Hi there,

I do need to have columns separated by some spaces. At this point I don't know how to achieve this…

Thanks for any help.

iRyusa:
Hello @EmmanuelBeziat 

Column doesn't support padding, you can just add the same padding on every element inside your column like this:

```xml
<mj-column>
  <mj-text padding="0 20px">...</mj-text>
  <mj-anything padding="0 20px">...</mj-anything>
</mj-column>
<mj-column>
  <mj-text padding="0 20px">...</mj-text>
  <mj-anything padding="0 20px">...</mj-anything>
</mj-column>
```

EmmanuelBeziat:
Hello,

Nope, this won't work for what I need here — really spaces between columns, not inside. Here's a scheme : 

![](http://image.noelshack.com/fichiers/2016/13/1459246934-example.jpg)

So at this point, each column contains two ```<mj-text>```and one ```<mj-image />```. Each '''<mj-text>''' have some padding on them already, and they only work "inside". In this case, i'm stucked because I have a need to have 3 columns with a visible gap between them. Needless to say, those columns must be responsive.

I was only able to achieve this by… Not using MJML for all this portion of the mail. This scheme is used 2 times in it, so basically using MJML was almost pointless for this project, I ended up with something like this : 

```markup
<mj-body>
  <mj-section>
    <mj-column>
      <mj-raw>
        <!-- some nasty, old, heavy, handwrited html -->
      </mj-raw>
    </mj-column>
  </mj-section>
    <mj-column>
      <mj-raw>
        <!-- some nasty, old, heavy, handwrited html -->
      </mj-raw>
    </mj-column>
  </mj-section>
</mj-body>
```

Moreover, it forced me to add some classes and CSS to handle responsiveness of this handwrited html, so each time I compiled using MJML, I was forced to add some CSS in the output version, because I couldn't do it inside the MJML template.

I had a dozen of mailing to do since the beginning of 2016, and most of them had some "plain-color" columns like this that need a gutter. This is why I think this would be really, really important to improve this system of columns, allowing some gap (maybe with an attribute ?) that would add some blank cells between actual ones.

Thanks for your help anyway :)

muxica:
Hey @EmmanuelBeziat, a temporary solution could be adding an empty column set to the specific width you need in between but this will probably result in strangely sized columns if you leave them at the default 100% width.
[...]

PR #241, opened 2016-05-10
Allow padding on mj-container component

### LINKED ISSUE #204: How do you set document/mj-body gutter ?
Hey guys,

Just started my first email with MJML which looks a good framework.
Otherwise, I'd like to set a left and right gutter outside my general template. In fact, i'd like to add a padding-left and padding-right to the body so my container has spaces on the sides.

group 3

I tried several things like putting in additional mj-column, settings padding on mj-body or wrapping with a mj-section but nothing works. Any help would be very appreciated ;)

Keep up the good work.

Here is my code : [http://codepen.io/Pixelize/pen/qZJWyd](http://codepen.io/Pixelize/pen/qZJWyd)

(sorry but github code tag is totally bugging)

pixelize:
Fix #204 #160

### DIFF
--- a/packages/mjml-container/src/index.js
+++ b/packages/mjml-container/src/index.js
@@ -66,7 +66,12 @@ class Container extends Component {
     return {
       div: {
         backgroundColor: mjAttribute('background-color'),
-        fontSize: defaultUnit(mjAttribute('font-size'), "px")
+        fontSize: defaultUnit(mjAttribute('font-size'), "px"),
+        padding: defaultUnit(mjAttribute('padding'), "px"),
+        paddingBottom: defaultUnit(mjAttribute('padding-bottom'), "px"),
+        paddingLeft: defaultUnit(mjAttribute('padding-left'), "px"),
+        paddingRight: defaultUnit(mjAttribute('padding-right'), "px"),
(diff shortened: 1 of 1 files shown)


iRyusa:
Hello @pixelize 
`padding` is supported only on `td` and `p` for every client that's we can't add it on the container
And it could break so many thing that being computed from container's width

pixelize:
Hey @iRyusa, it's not working on p or div on Outlook clients right ? Does it matter if we can apply it anyway and accept that will not not rendered in Outlook but in others email/web clients ?
Just did some test on Litmus, my test layout was ok on these ! Btw, you're more expert than me in css properties support in email clients, it's up to you 👍

ngarnier:
Hey @pixelize, we really appreciate your PR, thanks for contributing!

Unfortunately, as @iRyusa said, merging this would lead to too many uncontrolled issues. We want MJML to be responsive and supported on every major email clients, which includes Outlook, without requiring users to dig in what works and what doesn't for each email clients. 

I hope it makes sense to you and hope to have more contributions from your side, glad to have you onboard!

pixelize:
Happy to help such a great project !
No problem, I totally agree with you. Good to know of you want MJML to be driven, will keep in mind for next PR :)

PR #276, opened 2016-05-24
Removed the word facebook

jitendravyas:


ngarnier:
hey @jitendravyas, thanks for your PR but we'll keep the word "facebook" here, it shows that you can accumulate different social networks, including custom ones. Here, the example shows how to add Facebook + a custom social network.

PR #277, opened 2016-05-24
Fix mjml-section background-color attribute when background-url attribute is set.

Everus:


iRyusa:
Hey @Everus !

Thanks for your contribution 👍

PR #293, opened 2016-05-31
Add -t option in cli for setting up document title fix #283

### LINKED ISSUE #283: Support for custom title tags?
Apparently Android 4.4 is using `<title></title>` tags for push notifications. Also some clients like Outlook might allow you to view email in a browser for better rendering. In those cases title tag would be useful. At the moment I haven't figured out any way to add custom titles for templates.

pixelize:
Use case :

-  "View this email in your browser"
-                   Push notifications on Android

### DIFF
--- a/doc/install.md
+++ b/doc/install.md
@@ -54,3 +54,14 @@ If you like live-coding, you might want to use the `-w` option that enables you 
 It can be time-saving when you can just split you screen and see the HTML output modified when you modify your MJML.
 
 Of course, the `-w` option can be used with an `--output` option too.
+
+### Setting up document title
+
+```
+
+$> mjml -t My email
+
+
+```
+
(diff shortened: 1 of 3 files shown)


iRyusa:
Hi @pixelize as mentioned on Slack, 

I'd prefer to have a markup solution ( with `mj-title` in head ) instead of a cli solution. I am updating the internal API of `mj-head` to give more flexibility with the document. Does it make sense to you too ?

pixelize:
<mj-title /> would be awesome, couldn't figure it out how to make with the actual mj core system so I did like this :D

iRyusa:
Hi @pixelize `<mj-title>` will released this week, I close this PR as mentioned earlier. Thanks for your contribution

PR #294, opened 2016-05-31
<mj-include path='hello.mjml' />

GuillaumeBadi:


packages/mjml-core/src/includeExternal.js
```diff
@@ -0,0 +1,33 @@
...
+  const template = fs.readFileSync(path, 'utf8')
+  const content = getContent(template)
+
+  if (!content) { throw new Error(`Error while parsing file: ${path}`) }
+
+  return content
+})
+
+/*
+const ret = translate(`
```
iRyusa: to remove

GuillaumeBadi: ?

PR #321, opened 2016-06-28
MJML 3

iRyusa:
## Features

- `mjml-validator` provides validation to MJML documents with 3 levels ( skip/soft/strict)
- `mj-style` enables users to add CSS in the head that will be inlined in the generated HTML
- `mj-section` and `mj-column` now support the long awaited `border` (+top/left/right/bottom) attributes (`container` needs more work)
- `mj-social` now supports the `inner-padding` attribute to add padding between social networks
- `mj-font` now enables users to override default MJML fonts
- `mj-button` now supports the `border-top/right/bottom/left`, `width` and `height` attributes
- `mj-section` now support the `direction` attribute to change the way columns display on desktop

## Fixes
- Glob pattern for multiples files
- Fix recursive `mj-include` #337
- Now `.mjml` is truely optional for `mj-include`
- Fix path issue with `.mjmlconfig` (relative path should now work)
- `mj-group` now correctly supports `background-color` and should render well in outlook 2000/2003


iagomelanias:
Can the `mj-style` be used to style mjml tags, like `mjml-section`?

iRyusa:
Nop @iagomelanias you got `mj-attributes` for that with class support, override default values

iagomelanias:
Thanks @iRyusa. :)

The problem of the `mj-attributes` is the limitation of using only attributes available in the component. So, i can't "really" customize the component, a attribute for every css property is needed. And that makes the design possibilites so low. =/

Right now I have a prototype of a email design for my project and i can't make it in the mjml because there are some properties i can't use (like border for buttons, sections, etc). That is a big limitation.

Did you already thought about any feature that makes full css customization of the email available? I can contribute if possible.

iRyusa:
Well with mj style you can alter anything as a post render hook so you can add everything on final render. 

I guess that css selectors will be a little tricky but i think it can fit your need perfectly.

I wonder if we should set an `id` attributes for everything so it can be used in mj style ( it will be inlined at the end so all `ids` will be lost after mj-style is applied

iagomelanias:
Nice. I think this kind of feature is important to make more creative emails, out of defaults shown in the [templates](https://mjml.io/templates).

The `id` is a option, maybe even using the `tag` name is a good alternative, if possible. 
Also , the mj-style attribute for components is something really great, like the [ng-style](https://docs.angularjs.org/api/ng/directive/ngStyle), allowing to set the css properties from a json formatted string.

It allows the developers make more customized emails so easily. Would help a lot.

Thanks!

iRyusa:
The main issue with JSON notation is people from email development are not so familiar with JSON, so they'll need to learn it in top of MJML. I don't think it matches the goal of MJML to ease email development

Kerumen:
Nice! Very impatient for the border things. Do you have an ETA of merging and releasing this PR? :)

iRyusa:
We're removing XSD, but adding a new rule engine for validation !

 I don't know if it make sense to open the API to add custom validation rules

Issue #330, opened 2016-07-14
Button padding missing in Outlook


nsamala:
In version `2.3.0`, it looks like padding isn't being added to `<mj-button>`. 

On Gmail / Inbox, it renders fine:
![image](https://cloud.githubusercontent.com/assets/1831746/16848661/fff8c30e-49b3-11e6-80e2-b4ad527b76b1.png)

On Outlook, it used to render fine but now renders like this:
![image](https://cloud.githubusercontent.com/assets/1831746/16848695/39698baa-49b4-11e6-8e02-a120f0d262ee.png)

g0ddish:
It looks ok in 2016 but all other outlooks are broken, I will take a look.
Here is MJML to reliably reproduce this bug.

```
<mjml>
  <mj-body>
    <mj-container>
      <mj-section>
        <mj-column>

          <mj-button background-color="#F63A4D" inner-padding="50px" href="#">
            Promotion
          </mj-button>
        </mj-column>
      </mj-section>
    </mj-container>
  </mj-body>
</mjml>
```

g0ddish:
The padding is being applied to the anchor tag and not the parent td holding it, I swapped them and the button retains the padding on all clients (litmus).

PR #331 has a fix for this, thanks 😸

255kb:
Hi, thanks for the fix @g0ddish 
@iRyusa any ETA for a 2.3.1 mentioned in #331

iRyusa:
Should be around next week~

255kb:
ok thanks!

iRyusa:
2.3.1 has been released for a while now, but try it live on mjml.io is not up to date.

PR #331, opened 2016-07-14
Move internal padding to parent TD

g0ddish:
This should fix issue #330 I've tested this on litmus all clients with no issues please let me know if it causes unacceptable changes otherwise. A copy of the test MJML is in the ticket as well.
Thanks


iRyusa:
Oh... We've tested it without any href... so padding is applied on `p` but not on `a` nice catch, we will release a 2.3.1 with this !

lRlevolution:
The only problem with this is now that the padding is on the TD the actual a tag button radius is reduced to the text itself and not the padding around the a tag.

iRyusa:
Radius is applied on both `td` & `a/p` tag, so it shouldn't be an issue ?

PR #342, opened 2016-07-26
Fix mjml-column can deal with one children

martinhorsky:


iRyusa:
Hello thanks for your PR, 

I'm not fully understand why you need `React.map` ?

```
<mjml>
  <mj-body>
    <mj-section>
      <mj-column>
        <mj-text>Hello</mj-text>
      </mj-column>
    </mj-section>
  </mj-body>
</mjml>
```

martinhorsky:
Please see: https://facebook.github.io/react/tips/children-props-type.html

> Usually, a component's children (this.props.children) is an array of components

> However, when there is only a single child, this.props.children will be the single child component itself without the array wrapper. This saves an array allocation.

When you deal with React component children you need to use `React.Children` utilities.

https://facebook.github.io/react/docs/top-level-api.html#react.children

To fully understand, here is a snippet from my own component module:
```
  render() {
    const { mjContent } = this.props;

    const content = cleanContent(mjContent());

    return (
      <MJMLSection parentWidth={this.props.parentWidth} padding="0 10px">
        <MJMLColumn width="100%">
          <MJMLText
            color="#000000"
            font-family="Helvetica, Arial, sans-serif"
            font-size="24px"
            font-weight="bold"
            padding="20px 0 10px"
          >
            {content}
          </MJMLText>
        </MJMLColumn>
      </MJMLSection>
    );
  }
```

iRyusa:
In this case you're using `mjContent` that rely on that technique https://github.com/mjmlio/mjml/blob/master/packages/mjml-core/src/decorators/MJMLElement.js#L80

`props.children` is overriden by `buildProps` on mjElement too, so that shouldn't be an issue here ?

martinhorsky:
But issue is not related to `mjContent`. Let's say we have own simple component `<my-hello-world-text />` with source code similiar to mine:
```
render() {
    return (
      <MJMLSection>
        <MJMLColumn width="100%">
          <MJMLText font-weight="bold" color="#ff0000">
            Hello World!
          </MJMLText>
        </MJMLColumn>
      </MJMLSection>
    );
  }
```
Usage:
```
<mjml>
  <mj-container>
    <my-hello-world-text />
  </mj-container>
</mjml>
```
Error occurs `TypeError: children.map is not a function` 

Issue was introduced in commit https://github.com/mjmlio/mjml/commit/e31506c1e5c4b7c75f82ee9a3a586d53011ace07#diff-f05055ec67f20a0c84b13f69f0a3998cL128

---

Sure i can fix all my components and ensure having array inside `<MJMLColumn>`, but I think it's not best solution.

iRyusa:
I get it ! Thanks for the PR and for the explanation 👍

martinhorsky:
Thank you

PR #360, opened 2016-08-19
Expand MJML acronym in ReadMe to avoid ambiguity; add supporting image

Camcan:
Props for building such a brilliant module - this has made some traumatized friends of mine incredibly happy and there seems to be an overwhelming sense of relief that we don't have to write mind-numbing table-within-table html anymore when tasked with simple email templating work. 

"MJML" tho.
A month ago, after a bunch of discussion, we decided unanimously that MJML had to be short for Michael Jackson Markdown Language. We were always getting our tongues tied in trying to remember the acronym before this point, but after a tipsy jam to 'Thriller' the name Michael Jackson Markdown Language became unforgettable. 
Friday has rolled around again, and it's been a huge let-down for us to find the apparent extended title for this magical module is something as disappointing as Mail-Jet Markdown Language. I've made a quick fix to aid in fixing memorization of the module name in hope that others can have the same appreciation that we've had in the office for the past month.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,16 @@
 
 ---
 
+# Michael Jackson Markdown Language
+
+<p align="center">
+  <a href="http://mjml.io" target="_blank">
+    <img src="http://www.fanvasion.com/index/images/latestnews/michaeltribute.png">
+  </a>
+</p>
+
+---
+
 # Introduction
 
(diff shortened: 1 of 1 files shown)


iRyusa:
10/10 PR 👍

README.md
```diff
@@ -24,6 +24,16 @@
 
 ---
 
+# Michael Jackson Markdown Language
+
+<p align="center">
+  <a href="http://mjml.io" target="_blank">
```
ogonkov: https?

iRyusa:
(reopen because we've removed the current master sorry Michael ❤️ 🙈)

Issue #413, opened 2016-10-19
mjml-hero padding/height rendered as NaN


blaze33:
Using the mjml code for a fixed height hero from : https://mjml.io/documentation/?html#mjml-hero

MJML outputs:
 ```Warning: `NaN` is an invalid value for the `paddingTop` css style property. Check the render method of `Hero`.```

The resulting html contains:
```<td height="NaN" style="background-repeat:no-repeat;vertical-align:top;background-size:cover;background:#2a3448 url(https://cloud.githubusercontent.com/assets/1830348/15354890/1442159a-1cf0-11e6-92b1-b861dadf1750.jpg) no-repeat center center / cover;padding-top:NaN;padding-left:NaN;padding-right:NaN;padding-bottom:NaN;background-position:center center;" background="https://cloud.githubusercontent.com/assets/1830348/15354890/1442159a-1cf0-11e6-92b1-b861dadf1750.jpg">```

Issue may have been introduced by https://github.com/mjmlio/mjml/commit/8ae319aa53b33c4637497c529c8c0d3c4b896d43 in Hero.js by setting padding-* to null but I'm not really sure.

Quick test shows that [paddingParser](https://github.com/mjmlio/mjml/blob/master/packages/mjml-core/src/decorators/MJMLElement.js#L161) indeed try to `parseInt(null)` which result in NaN.

Not sure why this behaviour only happened on mjml-hero as other component also have padding-* to null... 

Quick fix is to check for both undefined and null values as follow:
```
-      if (typeof paddingDirection !== 'undefined') {
+      if (paddingDirection != null) {
```
paddingDirection is always defined so typeof is not needed to avoid a ReferenceError (http://stackoverflow.com/a/15992131/343834 helped me in this case).

I have a pull request coming.

PR #414, opened 2016-10-19
mjml-hero: fixed NaN padding

blaze33:
cf. issue https://github.com/mjmlio/mjml/issues/413

Code review welcome :)


iRyusa:
Hi !

It should be good! I'll review that tomorrow, I think it's not the only thing impacted by this kind of stuff. Adding default attribute to null with `mjml-validation` might have break some similar things.

Btw thanks for your implication in MJML ! It helps the community a lot !!

PR #439, opened 2016-11-15
Make mjml render in the browser.

### LINKED ISSUE #395: mj-font, mj-attributes or mj-title fail to render in the browser
Only just started playing so it's very likely I'm doing something wrong but when I try any of the mj-head components it fails to render.

```
import {mjml2html} from 'mjml';

const html = mjml2html(`
<mjml>
  <mj-head>
    <mj-font name="Anton" href="https://fonts.googleapis.com/css?family=Anton" />
  </mj-head>
  <mj-body>
    <mj-container>
      <mj-text font-family="Anton">Hello there!</mj-text>
    </mj-container>
  </mj-body>
</mjml>`);
```
Throws: Uncaught TypeError: Cannot read property 'name' of undefined

This is the same for mj-title and mj-attributes.

Changing how the attributes are accessed fixes it... feels like this can't be broken for everyone though - maybe it is?
```js
# from...
fonts.push({ name: el.attribs.name, url: el.attribs.href });

#to...
fonts.push({ name: el.getAttribute('name'), url: el.getAttribute('href') });
```

ianchadwick:
parseHead() now gets the element attributes using dom.getAttributes() and passes them to the handler in the same way as mjmlElementParser does. I've updated the attributes and font tags to use the newly passed attributes object.

Helps with #274 & fixes #395

### DIFF
--- a/packages/mjml-core/src/parsers/document.js
+++ b/packages/mjml-core/src/parsers/document.js
@@ -58,8 +58,14 @@ const mjmlElementParser = (elem, content) => {
 const parseHead = (head, attributes) => {
   const $container = dom.parseHTML(attributes.container)
 
-  each(compact(filter(dom.getChildren(head), child => child.tagName)), element => {
-    const handler = MJMLHeadElements[element.tagName.toLowerCase()]
+  each(compact(filter(dom.getChildren(head), child => child.tagName)), el => {
+    const element = {
+      attributes: dom.getAttributes(el),
+      children: el.children,
+      tagName: el.tagName.toLowerCase(),
+    };
+
+    const handler = MJMLHeadElements[element.tagName]
 
     if (handler) {
       handler(element, { $container, ...attributes })
--- a/packages/mjml-core/src/parsers/document.js
+++ b/packages/mjml-core/src/parsers/document.js
@@ -2,6 +2,7 @@ import { ParseError, EmptyMJMLError, NullElementError } from '../Error'
 import compact from 'lodash/compact'
 import dom from '../helpers/dom'
 import each from 'lodash/each'
+import toArray from 'lodash/toArray';
 import filter from 'lodash/filter'
 import { endingTags } from '../MJMLElementsCollection'
 import MJMLHeadElements from '../MJMLHead'
@@ -61,7 +62,7 @@ const parseHead = (head, attributes) => {
   each(compact(filter(dom.getChildren(head), child => child.tagName)), el => {
     const element = {
       attributes: dom.getAttributes(el),
-      children: el.children,
+      children: toArray(el.childNodes),
       tagName: el.tagName.toLowerCase(),
     };
 
--- a/packages/mjml-head-attributes/src/index.js
+++ b/packages/mjml-head-attributes/src/index.js
@@ -5,10 +5,10 @@ import omit from 'lodash/omit'
 
 export default {
   name: "mj-attributes",
-  handler: ($, { defaultAttributes, cssClasses }) => {
-    each(compact(filter($.children, child => child.tagName)), elem => {
-      const tagName = elem.tagName.toLowerCase()
-      const attributes = elem.attribs
+  handler: (element, { defaultAttributes, cssClasses }) => {
+    each(compact(filter(element.children, child => child.tagName)), elem => {
+      const tagName = element.tagName
+      const attributes = element.attributes;
 
       if (tagName === 'mj-class') {
         return cssClasses[attributes.name] = omit(attributes, ['name'])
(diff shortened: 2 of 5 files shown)


iRyusa:
Looks good ! Does it work in both node & client side ?

ianchadwick:
I've only tested a single large MJML file but that works in both - we're only just starting to use MJML so don't have a whole lot of examples to test against.

Among the standard layout tags it uses each of the head tags: attributes, font, style and title.

iRyusa:
Hello ! I tested your PR, and it looks good for the most part !

Only thing I can't get to work with jQuery is the mj-style / mj-title . For some reason we're filtering on child.type ` const innerText = el.children.map(child => child.type === 'text' && child.data).join('')` and it's undefined with jQuery. 

Do you have any idea on how to achieve that ?

ianchadwick:
I suck. So apparently I was lying when I said I was testing all 4 head components =/

I've committed a fix for the style and title. Feels like it might be time to make the dom.getChildren() method return a more normalised list of children for both Cheerio and jQuery. Unfortunately I'm a bit snowed under with work to do this right now.

For sanity, here's the MJML I used for testing both in node and from Chrome.

```xml
<mjml>
  <mj-head>
    <mj-title>Hello Kitty</mj-title>
    <mj-font name="'Source Sans Pro'" href="https://fonts.googleapis.com/css?family=Source+Sans+Pro" />
    <mj-attributes>
      <mj-all font-family="'Source Sans Pro'" />
      <mj-class name="element-text" font-size="16px" font-family="'Source Sans Pro'" />
      <mj-class name="element-header" font-size="20px" font-family="'Source Sans Pro'" />
      <mj-class name="green-button" font-size="15px" font-family="'Source Sans Pro'" background-color="#5db734" color="#FFFFFF" />
    </mj-attributes>
    <mj-style>
      .red-color {
        color: red;
      }
    </mj-style>
  </mj-head>
  <mj-body>
    <mj-container>
      <mj-section full-width="full-width" background-color="#FFFFFF">
        <mj-column>
          <mj-text mj-class="element-header" align="center">
            <h1 style="font-size: 28px">We need to talk...</h1>
          </mj-text>
          <mj-image src="https://images.pexels.com/photos/8923/pexels-photo.jpg?w=940&h=650&auto=compress&cs=tinysrgb" />
        </mj-column>
      </mj-section>
      <mj-section>
        <mj-column>
          <mj-text mj-class="element-text">
            <p>Cats have been domesticated for around 4,000 years. While they were once valued for their hunting abilities, they are now valued for their companionship and loving behaviour.</p>
            <p>While not well known, the collective nouns used for cats and kittens are a clowder of cats and a kindle of kittens.</p>
            <ul>
              <li class="red-color">Cats have 30 teeth (dogs have 42) and most of us know how sharp they are!</li>
            </ul>
            <p>Our domestic cats are known as little cats. They differ from large cats such as lions and tigers because they are naturally active at night and can purr.</p>
            <p>Cats are now the most popular pet in the UK and in the US.</p>
          </mj-text>
        </mj-column>
      </mj-section>
    </mj-container>
  </mj-body>
</mjml>
```

packages/mjml-core/src/parsers/document.js
```diff
@@ -58,8 +59,14 @@ const mjmlElementParser = (elem, content) => {
 const parseHead = (head, attributes) => {
   const $container = dom.parseHTML(attributes.container)
 
-  each(compact(filter(dom.getChildren(head), child => child.tagName)), element => {
-    const handler = MJMLHeadElements[element.tagName.toLowerCase()]
+  each(compact(filter(dom.getChildren(head), child => child.tagName)), el => {
+    const element = {
+      attributes: dom.getAttributes(el),
+      children: toArray(el.childNodes),
```
iRyusa: Shouldn't we call `dom.getChildren` and add the `toArray` inside `getChildren` for jQuery ?

iRyusa:
hi @ianchadwick !

I would like to merge it this week ! Can you take a look at my comment and give me your thoughts about it ? 

Thanks !

packages/mjml-core/src/parsers/document.js
ianchadwick: Hello! The problem with using dom.getChildren is that it returns element.children rather than element.childNodes and element.children doesn't contain text nodes. This breaks <title>I'm a text node</title>.

If changing dom.getChildren to return childNodes won't have any side effects then this would work. I'm off on holiday on the 1st so time is really limited. I'll try and find some time to at least do a test and compare the outputs using the original and then dom.getChildren with childNodes

iRyusa: Well for now i'm ok with that it works well ! Really appreciate the help ! Thanks and have a nice holiday then 👍

PR #444, opened 2016-11-22
Allow setting the height of a mjml text container

rogierslag:
This allows setting the height of a `<mj-text>` element. Especially for grid based layout this can be useful, such that each element in the grid has the same height, regardless of the number of characters in the `<mj-text>`

PR #453, opened 2016-11-28
Allow setting of the height of a mj-text

rogierslag:
This is a continuation of https://github.com/mjmlio/mjml/pull/444

It allows setting the height on a `<mj-text>` element. 

Litmus example (see the second column in the second row) https://litmus.com/builder/d15d2a8


iRyusa:
👍  for me

iRyusa:
@rogierslag can you fix conflicts please ? 👍 
I think it will be release in 3.1.1, we have to do some test internally with some of our templates

rogierslag:
Fixed

PR #456, opened 2016-12-02
Setting the process output to something different to 0 when the command fails

igalarzab:
Hi,

First of all, thank you for creating mjml. It's a wonderful tool.

I'm trying to use the CLI in a programmatically way but I'm having problems to validate if everything run ok or if there was any problem compiling my templates. The CLI is always returning 0 as exit code and all the output, even errors, are being written to the stdout (no stderr) so is very difficult to check how the tool did.

I added some changes to be able to rely on the process output of the CLI. As usual: 0 OK, something else KO.

What do you think?

### DIFF
--- a/packages/mjml-cli/bin/mjml
+++ b/packages/mjml-cli/bin/mjml
@@ -3,6 +3,8 @@
 var cli = require('../lib/client')
 var binary = require('commander')
 
+var commandResult = Promise.resolve()
+
 /*
  * Parse the command line arguments
  */
@@ -16,7 +18,7 @@ binary
       output: binary.output
     }
 
-    cli.renderFiles(Array.isArray(files) ? files.join(' ') : files, options)
+    commandResult = cli.renderFiles(Array.isArray(files) ? files.join(' ') : files, options)
   })
   .option('-l, --level [level]', 'Specifies the level of validation of MJML parser (skip/soft/strict)', /^(skip|soft|strict)$/i, 'soft')
   .option('-r, --render <file>', 'Compiles an MJML file')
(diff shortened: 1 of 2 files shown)


iRyusa:
Hi there @igalarzab and thanks for your contribution !

There's a big refactor of MJML cli coming soon, and I want to integrate your work in it !

I've changed the branch so we could get both refactor without loosing everything. Can you resolve your conflicts ? Sorry about the whole refactor on this file it's a really bad timing, and I can help you with the merge if you need anything feel free to reach me on MJML Slack !

igalarzab:
No problem :) I just did it. Please tell me if you find something wrong or have any ideas about how to improve it.

Also, I noted a bug in that branch, if you run something like:

```sh
$ mjml-cli first.html second.html
```

The parser doesn't work because the `glob()` doesn't find any file with the given pattern. I didn't fixed it because the array is joined by an space on purpose so I didn't know what was the original intention there.

Cheers

iRyusa:
Looks good for me !

Yeah I know about this one, array of file should be transpiled into a regexp to make `glob` happy. It's fixed in my local branch, but did commit most of it so you can resolve your conflicts 👍 

I'm also working on a better watch implementation with https://github.com/paulmillr/chokidar with mj-include support. I'm still figuring out if mj-include should be part of mjml cli rather than inside mjml-core... it should fix mj-include relative to the file that includes it rather than the current working directory. What do you think about it ?

igalarzab:
Probably I'm missing some context but IMO `mj-include` is a core component, I wouldn't move it to the CLI module. The CLI is just one gateway for running mjml. I see `mjml-cli` as a silly module, without any logic, with the only responsibility of getting input to be used by mjml and returning output to a CLI.

If the include directive is moved to `mjml-cli` it will not be possible to use it from outside of a CLI. Trying to compile a template in another kind of client (not a CLI) will fail because the component will not be defined.

But as I said, I'm probably lacking some context, so take my opinion with a pinch of salt :)

iRyusa:
While taking back a look at it, it means people that use gulp / grunt will not have access to `mj-include`. So you're right, it should stay inside mjml-core 👍

PR #473, opened 2016-12-13
Fix all linting errors

rogierslag:
The linting errors prevented my from committing easily, so I decided to get rid of them. Effectively there were three classes of errors

1. Escaped characters in character classes of regexes. Escapes there are not required, since characters are treated as literals.
1. Indenting
1. Import orders


gulpfile.babel.js
```diff
@@ -22,7 +22,7 @@ if (path.win32 === path) {
   srcEx = /(packages\\[^\\]+)\\src\\/
   libFragment = '$1\\lib\\'
 } else {
-  srcEx = /(packages\/[^\/]+)\/src\//
+  srcEx = /(packages\/[^/]+)\/src\//
```
rogierslag: since you are in a character class, the slash is not special but interpreted literally. The escape can be omitted

packages/mjml-core/src/helpers/postRender.js
```diff
@@ -1,4 +1,4 @@
-export const removeCDATA = str => str.replace(/<!--\[CDATA\[([^]*?)\]\]-->/gm, '$1')
+export const removeCDATA = str => str.replace(/<!--\[CDATA\[([^]*?)]]-->/gm, '$1')
```
rogierslag: Since you never actually opened the character class (due to the escape), the closing square bracket does not have a special meaning and therefore does not need escaping

rogierslag:
And a check to show everything should still behave correctly

https://regex101.com/r/zc9YH4/2
https://regex101.com/r/gpF7Pm/1

iRyusa:
Hi @rogierslag 
Thanks for your PR it looks good for me

rogierslag:
Seems like this was cherry picked into `master`?

iRyusa:
Mhh not really @rogierslag there's an issue while resolving conflicts ?

iRyusa:
Hi @rogierslag 

Some changes made it to master I think ( pretty sure someone fixed some in a PR ).
I think we can close this one from now with MJML 4 around the corner. Thanks !

PR #474, opened 2016-12-14
Fix escaping servers which replace the `&nbsp;` with a space

rogierslag:
We have seen some servers which replace the non-breaking space character with an actual space.
This should ensure that even when there is a space the div actually renders


rogierslag:
Also related to https://github.com/facebook/react/issues/1177 for anybody who uses mjml as a middle layer

PR #481, opened 2016-12-20
Mj Include ++

iRyusa:
- Watching nested Mj Include 
- Mj Include now path is relative to file
- Cli supports exit code now, thanks to @igalarzab
- Splitting some function into their own file for readability 
- Major cleaning in client legacy code


lohek:
👌

Issue #509, opened 2017-01-20
Cannot use mj-include inside mj-head


vbraun:
Starting with mjml 3.2.0, using mj-include inside mj-head does not work any more. We have a common <mj-attributes> section for multiple emails, each email including it as in
~~~~
<?xml version="1.0" encoding="utf-8"?>
<mjml>
    <mj-head>
         <mj-include path="common-mj-attributes.mjml" />
    </mj-head>
[...]
~~~~
Build does not error out, but prints the somewhat cryptic 
~~~~
Warning: No handler found for: mj-raw, in mj-head, skipping it
~~~~
message and does not include the attributes. This worked with mjml 3.1.1

iRyusa:
Hi @vbraun 

Because we change how that works in 3.2 ! It's weird that it mention `mj-raw` tho..

So now you can just put your mj-attributes inside a mj-head for you common-mj-attributes.mjml like that : 
```xml
<mjml>
  <mj-head>
    // Every head attributes will be includes in main file
   </mj-head>
</mjml>
```

vbraun:
That didn't work for me either. I debugged this a bit and it seems to be because we use mjml2html() in code to render the mjml. Then, in `includeExternal.js`, the `var filePath = _ref.filePath;` is undefined. Then, a few lines further down, `_path2.default.dirname(filePath || process.cwd()` is the parent directory of where we launch from. So all `path=` arguments are not relative to the current directory, but the parent directory of the current directory (this is most likely not intentional).

The second bug is then that `<mj-include path=...` read errors are not errored out (imho exit with code > 0 would be the right thing to do) but replaced with a `'<mj-raw><!-- mj-include: file not found ' + filePath + ' --></mj-raw>';`. Hence the uninformative `Warning: No handler found for: mj-raw, in mj-head, skipping it`.

If I make the `<mj-include path=...` relative to the parent of the current directory then it works, but that can't be the right way of doing things.

iRyusa:
Wait, how do you compile your file with the cli ? The cli should set filePath for you, that's an option that you can pass to mjml2html & mjmlrenderer too.

Ok so now I get it why'd you get a mj-raw inside mj-head ! I add this to add a comment inside mj-body so you do

vbraun:
Just to clarify, I do not use the cli. I'm using my own script that does `import {mjml2html} from 'mjml'` and then calls `mjml2html` in code.

vbraun:
I made #510 to clarify the parent directory issue

[...]

PR #510, opened 2017-01-20
Do not use parent directory of cwd

vbraun:
When filePath is undefined, use process.cwd() instead of its parent
directory. See #509


iRyusa:
Thanks @vbraun for the fix !

PR #511, opened 2017-01-24
Allow disabling of blue links in iOS

rogierslag:
Whenever we send email containing user generated data, sometimes text and links may be colored blue on iOS.

Usually we do not want this to happen. Previously we added the following css to the head of the email to circumvent this

```css
a[x-apple-data-detectors] {
    color: inherit !important;
    text-decoration: none !important;
    font-size: inherit !important;
    font-family: inherit !important;
    font-weight: inherit !important;
    line-height: inherit !important;
}
```

As a solution I've added this flag to the `mj-text` component. If the flag is set, iOS will not try to style these automatically generated links

Litmus example (note how the magnet.me in the text is not autolinked, whereas I let it autolink near the footer

https://litmus.com/builder/b0b07bc

### DIFF
--- a/packages/mjml-core/src/configs/defaultStyle.js
+++ b/packages/mjml-core/src/configs/defaultStyle.js
@@ -7,4 +7,12 @@ export default `
   table, td { border-collapse:collapse; mso-table-lspace: 0pt; mso-table-rspace: 0pt; }
   img { border: 0; height: auto; line-height: 100%; outline: none; text-decoration: none; -ms-interpolation-mode: bicubic; }
   p { display: block; margin: 13px 0; }
+  .no-blue-link-ios a[x-apple-data-detectors] {
+    color: inherit !important;
+    text-decoration: none !important;
+    font-size: inherit !important;
+    font-family: inherit !important;
+    font-weight: inherit !important;
+    line-height: inherit !important;
+}
 `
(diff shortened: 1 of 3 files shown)


dalefish:
Nice

Kind Regards,

:----------------------------:
Dale McConnell

iRyusa:
Hi @rogierslag and thanks again for your contribution

Sorry for the delay, so I like the snippet to disable blue links in iOS but I think it shouldn't be an attribute for any mj-element. So every component that will possibly have any text inside it would require to have this attribute.

I think you can safely add a easy class name that we can document and people only have to apply it to an html block like `span/p/div`

What do you think about it ?

rogierslag:
That would be possible as well. Would you in that case still add the default styling to the head of the email, or should people do that themselves as well?

I think that in order to keep mjml easy to use, and ready to go such snippets should be present. If not, it would encourage people to look around for solutions, which is what MJML tries to offer

iRyusa:
Hi @rogierslag 

I think this is totally something that you want to offer inside MJML. We could keep the css and add some doc about the class !

rogierslag:
Ok, will do a fixup somewhere today :)

iRyusa:
Hi @rogierslag !

Any news on this ? 👍

iRyusa:
Hi there,
A
ny news on that ?

We might need 2 version of this : one with `.disable-ios-blue-links` and `.disable-ios-blue-links a` so it can be used with `css-class` attribute easily

iRyusa:
Hi @rogierslag 

We're closing this PR because we won't accept any new PR on non breaking stuff for MJML3. 

However you can still push this on the MJML4 branch `next` with this in mind : https://github.com/mjmlio/mjml/pull/511#issuecomment-324050223

PR #514, opened 2017-01-25
Gave that error message more meaning

michelgotta:
If \<mj-body\> contains just text the element is not "empty" but the parsing still fails.


iRyusa:
Hi @michelgotta 

Thanks for this contribution 👍

PR #519, opened 2017-01-26
Full JSON support

iRyusa:
Make JSON great again


packages/mjml-core/src/parsers/document.js
```diff
@@ -109,11 +101,17 @@ const documentParser = (content, attributes) => {
     throw new EmptyMJMLError('No root "<mjml>" or "<mj-body>" found in the file, or "<mj-body>" is empty')
   }
 
-  if (head && head.length === 1) {
-    parseHead(head.get(0), attributes)
+  return {
```
iRyusa: We might want to do a generic parser for both head & body. But with MJML 4 coming, I think it should be ok for now to keep them split because both head/body components are still registered in their own store

PR #523, opened 2017-01-26
Better Output Logic

adamyi:
This patch fixes two things:

1. Originally, `-o` and `-s` cannot be used simultaneously;
2. Originally, if we specify `-i` without `-o` or `-s`, it will run into errors.

With this patch, it will first check that, if there's no filename and no output stream specified, it will automatically output to stdout. After that, it will check filename and stdout separately.

### DIFF
--- a/packages/mjml-cli/src/client.js
+++ b/packages/mjml-cli/src/client.js
@@ -51,10 +51,15 @@ const render = (bufferPromise, { min, output, stdout, fileName, level }) => {
         handleError(availableErrorOutputFormat['text'](errors))
       }
 
-      if (stdout || (!output)) {
+      if (!(stdout || output)) {
         process.stdout.write(html)
       } else {
-        return write(output, html)
+        if (stdout) {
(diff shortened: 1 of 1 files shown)


iRyusa:
Hi @adamyi and thanks for your contribution !

I have a few issues with your PR : 

> Originally, -o and -s cannot be used simultaneously;

Is this something that we'd want ? I mean is there any use cases ? 

> Originally, if we specify -i without -o or -s, it will run into errors.

I think it should be specific from the `-i` option, I would like to get this logic moved out inside `mjml-cli/bin/mjml` rather to this and throwing an error if both output/stdout option are empty

What do you think about it ?

iRyusa:
Hi @adamyi !

I'd like to know if you're still on this patch ! Thanks !

iRyusa:
Hi @adamyi 

With no news from you i'm closing this PR then, we'd be happy to reopen if you're still working on this ! Thanks

PR #532, opened 2017-02-02
allow adding class attribute to some elements

### LINKED ISSUE #480: Add class to element
I would like to use classes in my mjml, that are also transpiled as class in the resulting html.

e.g. `<mj-text class="viewbrowser>` -->  `<div class="viewbrowser>`

Adding a class just to `<mj-section>` would also be sufficient for me.
Use case: The newsletters show a link to a browser view, which is useful if the mail client doesn't display it properly. The browser view should hide this link.
The mailserver detects this by adding a class "viewbrowser" to the element.

This would save me some time of manually editing every newsletter.

adekbadek:
Allows class attribute in mj-text, mj-section, mj-container, mj-column, mj-image.
Resolves #480


iRyusa:
Hi @adekbadek and thanks for this PR !

We're discussing about this internally and the best way to add this into MJML.
Something I have in mind is the `class` attribute name. We already have `mj-class` what do you guys thinks about prefixing by `css-`. 

I'm not sure about `mj-container` getting additional `css` classes, you can only have one per email so I don't think this add much.

Scritik:
I vote for `mj-class`

When we add the attribute, it will add a class, and if some stuff is in the head relating to this class, it will added too. This way, you keep it simple.

pixelize:
I agree. I think we need to have 2 classes to avoid miss reading. `mj-class` for mj-elements as we use it actually and the other could could `css-class` for managing mediaqueries ie. as you suggested.

iRyusa:
Hi @adekbadek 

We want to merge this for next 3.3.1 but it still need a little work on it. So as we discussed, we want to change `class` to `css-class` :  it should avoid confusion about `mj-class` and regular css classes. 

I think we need to add this to he MJMLElement decorator root `td` for every elements that lives inside a column can now support `css-class`. CSS classes should be applied on root tag of each elements, and add the same class with a `-outlook` suffix on potential outlook conditionnal comments.

What do you think about it, do you have a little time to work on this ?

Thanks a lot for your work !

adekbadek:
Hi @iRyusa! 

I changed `class` to `css-class`. 

However I'm not so familiar with the project to understand what changes does `MJMLElement ` decorator need (perhaps you could point me to a readme or something?). 
I'll look into it, but in the meantime if someone wants to complete this PR they're welcome

PickRelated:
Hey guys! Thanks for your work! Any updates on this?

iRyusa:
We're working on improving this PR to ship it with 3.3.2 ! Sorry for not providing enough feedback on this :)

PickRelated:
Lol no sorries. You're giving a high-quality service like this is a paid software. Nice work, man! Thanks for your response.

RDeluxe:
Hi guys,

Just wondering ; so the goal here would also to have a check on the attributes passed to the mj element ? So that we can have errors displayed when we try to apply a class with non-supported attributes ?

ngarnier:
@RDeluxe, no, no check on the attributes. The idea is to pass a class name from a MJML element to the first HTML element generated from that MJML element. 

Example: if you add `css-class="myClass"` to a `mj-image`, the first HTML element generated from `mj-image` (`tr`) will bear `class="myClass"`. This will enable you to target your image in a custom CSS style more easily, especially useful for media queries!

iRyusa:
Hi @adekbadek 

I'm merging this in a branch of MJML repo to work on this. I will ping you when this will be ready to release ! Thanks for you work on this 👍

PR #534, opened 2017-02-03
Multiple fixes for 3.3 release

iRyusa:
- Mj-include in head
- Cli crash with validation skip
- Mj-Head with single element now being parsed correctly
- Avoid juice when no inlinable CSS


lohek:
👌

Issue #319, opened 2016-06-28
Button component line-height option


blackfyre:
Hi!

I've got some issues with the button component: https://www.dropbox.com/s/cxzrg8i88j02fhh/mjml-button-line-height-error.jpg

Basically in Outlook 2003 / 2017 / 2013 multi line buttons have to much for line height. It doesn't seem to be explicitly set, but an attribute for this would be great :smile:

ngarnier:
Hi @blackfyre, thanks for creating this issue. Indeed multi-lines buttons can be tricky, we'll investigate if we can safely introduce line-height to the button component (I'm pretty confident it won't cause any major issue).

blaugueux:
Cf. https://www.campaignmonitor.com/css/ ```line-height``` seems to be supported across almost every email client

eikaramba:
+1 for this. Also have problems because my button should display a long link that is breaking and now looks quite ugly because of the inner background-color of the a tag. Maybe another option would be to customize the inner background-color. Or set line-height.

ngarnier:
Fixed in 3.3.3: https://github.com/mjmlio/mjml/releases/tag/3.3.3

PR #537, opened 2017-02-06
Increase the line height in mj-button so charachters are not cut off on multiline buttons

rogierslag:
We just received work that on some email clients the text on buttons would be slightly cut off

example

Since this obviously looks bad, I quickly dove into this, and as it turns out a slight additional line-height is required (font-size and line-height do not match up exactly). By increasing this from 100% to 120% the issues are resolved (under multiple fonts)

Litmus test: https://litmus.com/checklist/emails/public/799bfc8

### DIFF
--- a/packages/mjml-button/README.md
+++ b/packages/mjml-button/README.md
@@ -42,6 +42,7 @@ font-style                  | string      | normal/italic/oblique               
 font-size                   | px          | text size                                        | 13px
 font-weight                 | number      | text thickness                                   | bold
 font-family                 | string      | font name                                        | Ubuntu, Helvetica, Arial, sans-serif
+line-height                 | px or %     | line height of the text in the button            | 120%
 color                       | color       | text color                                       | #ffffff
 text-decoration             | string      | underline/overline/none                          | none
 text-transform              | string      | capitalize/uppercase/lowercase                   | none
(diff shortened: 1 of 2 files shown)


rogierslag:
References https://github.com/mjmlio/mjml/issues/319

iRyusa:
Hi @rogierslag 

Sorry for the wait, can you provide a quick MJML to reproduce the issue with multiline button ?

The PR looks fine to me should we add a `mj-attribute` for `line-height` tho ?

rogierslag:
No problem.

You can check an example of the mjml at https://mjml.io/try-it-live/Syb7kQQFl

I agree with your suggestion, lets make the line-height customisable (since it also depends on how the font works exactly). A one size fits all will probably conflict for some users. As a default value, I suggest to keep it 120% to allow for a bit more spacing (this also makes the button better tappable on mobile devices)

rogierslag:
Build failure on an unrelated dangling comma

iRyusa:
Hi @rogierslag

Coming back to this, we're testing the new `mj-button` in #557 and if it's still an issue we'll ping back here to know if you can update the PR for it.

rogierslag:
👍

iRyusa:
Hi @rogierslag 
Fix is still needed, however we don't want to force any default unit on `line-height`. Would you want to update your PR or I can apply this PR #658 and credit you for the line-height in the changelog ?

rogierslag:
The latter option was fine! Sorry, but swamped with work :'(

PR #571, opened 2017-03-01
Clearify <mj-table/> documentation

bassbump:
<mj-table> does not allow child components, wasn't entirely clear from docs


ngarnier:
Thanks a lot @bassbump!

Issue #600, opened 2017-03-16
mj-raw not placed at the right position


Almya:
As said in the title, I want to mix mjml with a rendering template but the <mj-raw> tag does not position well. It seems that it doesn't wrap the commented tags well.

If I try to wrap a section with a twig tag

```
<mjml>
    <mj-body>
        <mj-container>
            <mj-raw>
                {% if online == false %}
            </mj-raw>
            <mj-section></mj-section>
            <mj-raw>
                {% endif %}
            </mj-raw>
        </mj-container>
    </mj-body>
</mjml>
```
the result is

```
<body>
  <div>
    <!--[if mso | IE]>
    <table role="presentation" border="0" cellpadding="0" cellspacing="0" width="600" align="center" style="width:600px;">
      <tr>
        <td style="line-height:0px;font-size:0px;mso-line-height-rule:exactly;">
          <![endif]-->
          {% if online == false %}
          <div style="margin:0px auto;max-width:600px;">
            <table role="presentation" cellpadding="0" cellspacing="0" style="font-size:0px;width:100%;" align="center" border="0">
              <tbody>
                <tr>
                  <td style="text-align:center;vertical-align:top;direction:ltr;font-size:0px;padding:20px 0px;"></td>
                </tr>
              </tbody>
            </table>
          </div>
          {% endif %}
          <!--[if mso | IE]>
        </td>
      </tr>
    </table>
    <![endif]-->
  </div>
</body>
```
But the result should be

```
<body>
  <div>
    {% if online == false %}
    <!--[if mso | IE]>
    <table role="presentation" border="0" cellpadding="0" cellspacing="0" width="600" align="center" style="width:600px;">
      <tr>
        <td style="line-height:0px;font-size:0px;mso-line-height-rule:exactly;">
          <![endif]-->
          <div style="margin:0px auto;max-width:600px;">
            <table role="presentation" cellpadding="0" cellspacing="0" style="font-size:0px;width:100%;" align="center" border="0">
              <tbody>
                <tr>
                  <td style="text-align:center;vertical-align:top;direction:ltr;font-size:0px;padding:20px 0px;"></td>
                </tr>
              </tbody>
            </table>
          </div>
          <!--[if mso | IE]>
        </td>
      </tr>
    </table>
    <![endif]-->
    {% endif %}
  </div>
</body>
```
Same if I try to use a for loop between columns

```
<mjml>
    <mj-body>
        <mj-container>
            <mj-section>
                <mj-raw>
[...]

PR #607, opened 2017-03-21
Replace problematic chars with placeholders 

iRyusa:
Avoid Juice to goes on infinite loop, and Cheerio to misinterpret some tag-like chars

### DIFF
--- a/packages/mjml-accordion/yarn.lock
+++ b/packages/mjml-accordion/yarn.lock
@@ -648,9 +648,9 @@ minimist@0.0.8:
   version "0.0.8"
   resolved "https://registry.yarnpkg.com/minimist/-/minimist-0.0.8.tgz#857fcabfc3397d2625b8228262e86aa7a011b05d"
 
-mjml-core@~3.3.0-beta.8:
-  version "3.3.0-beta.8"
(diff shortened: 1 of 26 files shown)


iRyusa:
Will probably paste this in MJML4 if you're ok with this @lohek

web-padawan:
Hi! Just curious: should this also allow using [SparkPost](https://developers.sparkpost.com/api/substitutions-reference.html#header-basic-substitution) templates? Is there any ETA on when will this be released?

ngarnier:
hey @web-padawan, yes SparkPost templating syntax should work fine with this fix (let us know anything you find!). We'll release 3.3.1 today :)

ngarnier:
Actually, due to an issue with NPM we can't publish today, sorry. Likely next week!

web-padawan:
@ngarnier Thanks!

web-padawan:
I've just upgraded to `3.3.2` and code like `{{if user.language=="ru"}}` or `{{each products}}` in text still gets stripped out when compiling, while inside of attributes it does not. 
Could you check that please?

iRyusa:
Can you provide some code to reproduce please ?

ngarnier:
Hey @web-padawan, you still need to add your templating snippets inside components such as `mj-text` or `mj-raw`, I think that's the issue.

web-padawan:
UPD: looks like fixing #600 will resolve my issue.

PR #643, opened 2017-04-18
Keep mj-group pixel widths as pixels

### LINKED ISSUE #641: Group px width gets converted to percentages
Noticed when I use a `mj-group` component and specify a pixel width, the output class is a calculated percentage instead of the provided pixel width.

adambullmer:
Fixes #641


iRyusa:
Hi @adambullmer and thanks for the PR

Can you fix the linting issue please ?

iRyusa:
Thanks again on this PR but as said in the #641 we're supporting only `%` in `mj-group` to avoid breaking mobile view for some device.

PR #663, opened 2017-05-04
Fully clickable button

iRyusa:
From @guillaumebreux original PR


bago:
Are you sure this works on Outlook 2013/2016 windows and Windows 10 Mail?
AFAICT links wrapping table won't work in the "word engine".

ngarnier:
Hey @bago, thanks for reporting! We did extensive tests in terms of rendering but couldn't test the actual click in all email clients. Do you have any way to test in those email clients?

madonkey:
Can confirm. Button isn't clickable on Outlook 2007.

bago:
AFAIK if you put one A outside the table and one A inside there will be other clients breaking anyway. I tried in past creating a "full clickable button" and I decided it was OK not to have fully clickable buttons :-)

You can test in Windows platform using Outlook 2007/2010/2013/2016 or Windows 10 Mail application.

For outlook you could put the external A in exclusive conditional comments and the internal A in inclusive conditional comments, but you'll find other (non outlook) webmails that will break the external A trying to close it before the <table> opening in order to sanitize it, so it won't work anyway.

So, my suggestion is to not change the button unless you are ready to do a lot of real world testing and A wrapping TABLES are not supported in a lot of weird ways.

bago:
You can find other people spending a lot of times on buttons:
http://blog.edmdesigner.com/bulletproof-html-email-buttons-a-new-line-height-based-approach/

And you can see there the 2 options are:
1) padding based: they looks good but they are not fully clickable
2) line-height based: they are fully clickable but they don't look good on outlook (vertical alignment).

You can use conditional comments to show one button style to word-based outlooks and the line-height solution to other clients, but you will have to do extensive real world testing (also on Windows 10 smartphones, now using Word rendering, too and being touch).

PR #668, opened 2017-05-10
3.3.3-beta 2 : Outlook Word button fix

iRyusa:
Try to fix this https://github.com/mjmlio/mjml/commit/219b32de8729df72e9f20fb681b5b273816263ee#commitcomment-22016570

### DIFF
--- a/packages/mjml-accordion/yarn.lock
+++ b/packages/mjml-accordion/yarn.lock
@@ -648,9 +648,9 @@ minimist@0.0.8:
   version "0.0.8"
   resolved "https://registry.yarnpkg.com/minimist/-/minimist-0.0.8.tgz#857fcabfc3397d2625b8228262e86aa7a011b05d"
 
-mjml-core@~3.3.0:
-  version "3.3.0"
-  resolved "https://registry.yarnpkg.com/mjml-core/-/mjml-core-3.3.0.tgz#4033694b2ac7a27efe387b72cb4d89ccc9886edb"
+mjml-core@~3.3.3-beta.1:
+  version "3.3.3-beta.1"
+  resolved "https://registry.yarnpkg.com/mjml-core/-/mjml-core-3.3.3-beta.1.tgz#a03f6e857054d1020fc843c316afc07f99b5a35b"
   dependencies:
     cheerio "^0.22.0"
     classnames "^2.2.5"
--- a/packages/mjml-button/src/index.js
+++ b/packages/mjml-button/src/index.js
@@ -89,7 +89,8 @@ const baseStyles = {
 const postRender = ($) => {
   $('.outlook-button-fix').each(function () {
     const $a = $(this)
-    const attributes = ['style', 'rel', 'href', 'target'].map(a => `${a}="${$a.attr(a)}"`)
+    const attributes = ['style', 'rel', 'href', 'target'].filter(a => $a.attr(a))
+                                                         .map(a => `${a}="${$a.attr(a)}"`)
                                                          .join(' ')
 
     $a.find('p').each(function () {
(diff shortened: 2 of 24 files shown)


bago:
As I already commented here: https://github.com/mjmlio/mjml/pull/663
I'm almost sure there are webmails out there that will try to well-format (as per HTML4) your message and will close your <a> tag before the opening of the <table> breaking the clickability of your button.

So your workaround will fix outlook (word-based) and you can adapt it to fix also old IEs in quirks mode (e.g: Outlook Express and old Lotus Notes/Windows Mail), but there is no workaround for webmails or anything else trying to sanitize your html code to html4 specifications.

Unfortunately I don't remember what webmail/s do that, but I prefer to alert you anyway so you will know you have to test real webmails because rendering tools won't show you this issue.

ngarnier:
Hey @bago, thanks a lot for sharing your knowledge on this, much appreciated! 

This fix was intended to fix the "look" of the button on outlooks but now we need indeed to try it in real-world situation. It's challenging but we'd really like to find a way to have fully clickable button. However, we definitely won't release this if it means the button is not clickable on popular email clients like the ones you mentioned.

bago:
OK, I found my old tests and here is the list of clients that will break a link when they find a table:
- Outlook 2007/2016 (word based)
- Windows 10 Mail
- RoundCube (used by OVH and others)
- T-online.de webmail
- Terra webmail (biggest webmail in america latina)
- Telstra/BigPond (biggest webmail in australia)

Word based clients (Outlook/Windows 10 Mail) can be fixed using conditional comments, but the others are a dead-end (at least in my tests).

Maybe a line-height based solution not involving a table will be a better foundation but I never investigated more. I'm just using a few pixel padded table and a line-heighted link inside.. it is not "fully clickable" but it works find across clients and most people clic on the text anyway (and touch devices are more forgiving).

BTW I share my knowledge so you start your investigation with a better perspective and if you find a better solution working everywhere, everyone will benefit, me included.

ngarnier:
Thanks again for this added information @bago. 

Update: after some tests, it seems that among email clients that are likely to cause issues, results are:
- 👌  on Outlook/Windows 10 (thanks to conditional comments) (text clickable)
- 👌  on the native Android app (fully clickable)
- 🚫  on webmails (tested with Roundcube only, for now) (not clickable at all)

ngarnier:
Update: as Webmails like Roundcube wouldn't like this option because of the conditional comment, we tried to simply remove the comment and have a layout with `a --> table --> a`. 

While it seemed to be working fine on most email clients, including mobile, desktop and web clients, Outlook 2007 doesn't like it and the button isn't clickable at all (text is selected instead). 

As we can't find any way to make the clickable button work *everywhere*, it is very likely that we'll abandon and keep the button as it is today.

iRyusa:
Pinging @guillaumebreux on this discussion, the main contributor of the fully clickable button

guillaumebreux:
:o What a bad news! 
But I won't giveup on that, I'm confident I can have something working (at least for my own prerequisites).

I'll let you know :)

bago:
PS, about the "on the native Android app (fully clickable)" I have to add that I just re-tested on a Nexus 7 stock Mail app (version 7.0) with Android 5.1 (not Gmail, the stock mail app) and the button is not clickable) Which phone/android version/app version did you test when you verified it was clicable? (se we at least fill some details)

ngarnier:
Hey @bago, the versions used were Android 6.0.1 and Mail app 3.3.78

bago:
Weird number for the stock email app. Are you sure you are on a stock mail app and not samsung mail? Which Phone?

I tested 6.x and 7.x:
http://www.apkmirror.com/apk/google-inc/email/

ngarnier:
Just checked with the person who tested and indeed, it was actually tested on Android 6.0.1 and Samsung Mail 3.3.78. We didn't investigate further as it wasn't a route to follow anyway, but you're right, it's good to know all the details for potential later explorations.

PR #682, opened 2017-05-18
Revert fully clickable button

iRyusa:


10kc-zack:
@iRyusa Any reason as to why this was removed?

iRyusa:
Hi @10kc-zack 
You can find more information about it here : #663 and here #668

10kc-zack:
@iRyusa Thanks for pointing me in the right direction!

PR #685, opened 2017-05-22
Validation tweaks

adekbadek:
in CLI, the `-l skip` option resulted in `errors` being undefined

### DIFF
--- a/packages/mjml-cli/src/client.js
+++ b/packages/mjml-cli/src/client.js
@@ -46,9 +46,10 @@ const render = (bufferPromise, { min, output, stdout, fileName, level }) => {
     .then(result => {
 
       const { html, errors } = result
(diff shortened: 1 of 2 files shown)


packages/mjml-cli/src/client.js
```diff
@@ -46,9 +46,10 @@ const render = (bufferPromise, { min, output, stdout, fileName, level }) => {
     .then(result => {
 
       const { html, errors } = result
+      const skipValidation = level === 'skip'
```
iRyusa: I don't really understand the reason behind that. Could you explain a bit please ? If level is skip, mjml-core send a empty array.

adekbadek: Tested it on a different machine and it works fine. Sorry for the hasty PR.

Issue #700, opened 2017-06-01
Upgrade deprecated fs-promise


mikelambert:
Installing mjml triggers the following warning:
`npm WARN deprecated fs-promise@2.0.3: Use mz or fs-extra^3.0 with Promise Support`

This is the file/line in question that needs to be updated:
https://github.com/mjmlio/mjml/blob/master/packages/mjml-cli/package.json#L21

ngarnier:
Released with [3.3.4](https://github.com/mjmlio/mjml/milestone/12).

PR #707, opened 2017-06-07
Replace deprecated fs-promise with mz

jackytck:
#700 Upgrade deprecated fs-promise with mz


iRyusa:
Hi @jackytck !
Thanks for you contribution ! And it looks ok for me, we can merge this for next release I think

PR #716, opened 2017-06-13
Adds srcset for the loading of responsive images.

ctborg:


iRyusa:
Hi @ctborg and thanks for your contribution !

Do you have any idea how well is supported srcset yet ?

ctborg:
Pretty much everywhere but IE.  Edge has some rendering issues.

http://caniuse.com/#feat=srcset

Having said that, most web based clients strip out that attribute. gmail, etc.  I feel it still has value for native mail clients though.

dalefish:
I had thought about using something like an attribute for `mobile-src=...`
then tying that together with a media query substitution where media
queries are supported? Or inverted (sub in the desktop image)?

Personally I prefer to go with subbing the mobile image in it means you
don't have to worry as much about things like outlook but it's not 100%
aligned with the mobile first methodology.



Kind Regards,

:----------------------------:
Dale McConnell

ngarnier:
I've seen it being discussed before, [HTeuMeuLeu did some research on it (in French)](https://www.hteumeuleu.fr/attribut-srcset-images-responsive/) and thinks [it can be used quite safely.](https://litmus.com/community/discussions/6152-srcset-and-sizes-in-email#comment-10373). 

@ctborg would you be able to provide Litmus tests of your PR in situation? You can use [the image part of the styleguide we use for testing](https://github.com/mjmlio/mjml/blob/master/packages/mjml/styleguide.mjml#L324).

ctborg:
@ngarnier: sure, is there a Litmus account I can use?

ngarnier:
We don't have any account to provide no, we'll do the tests then. I'm pretty confident anyway as it's a progressive enhancement: use `src` by default and use `srcset` for the clients that support it.

PR #728, opened 2017-06-23
Update readme to better reflect the output of mjml2html

jamesblight:
The documentation doesn't reflect that the output of mjml2html is an object and not a string of html.
I think it's useful to know that the output of mjml2html is an object that contains errors as well as the html.


iRyusa:
Hi @jamesblight 
Thanks for this PR  👍

PR #751, opened 2017-07-21
Add thin arrows to problematic chars

jayjun:
Elixir (therefore, EEx) uses thin arrows extensively.

```eex
<%= for item <- items do %>
  <mj-text></mj-text>
<% end %>
```

```eex
<%= case var do %>
  <% nil -> %>
    <mj-text></mj-text>
  <% var -> %>
    <mj-text></mj-text>
<% end %>
```

### DIFF
--- a/packages/mjml-core/src/MJMLRenderer.js
+++ b/packages/mjml-core/src/MJMLRenderer.js
@@ -23,7 +23,7 @@ import React from 'react'
 import ReactDOMServer from 'react-dom/server'
 import warning from 'warning'
 
-const DANGEROUS_CHARS = ['{{', '}}', '<%', '%>', '<=', '=>', '{%', '%}', '{{{', '}}}']
+const DANGEROUS_CHARS = ['{{', '}}', '<%', '%>', '<=', '=>', '<-', '->', '{%', '%}', '{{{', '}}}']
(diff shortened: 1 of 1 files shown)


iRyusa:
Hi @jayjun 
Thanks for this PR !

witzig:
Hello @jayjun and @iRyusa 

Adding the thin right arrow to `DANGEROUS_CHARS` has an undesired side effect as it breaks HTML comments and therefore the postrender functions. Given this MJML document:

``` xml
<mjml>
    <mj-body>
        <mj-container background-color="#eee">
            <mj-raw>
                <!-- some comment -->
            </mj-raw>
            <mj-section>
                <mj-column>
                    <mj-button >
                        Don't click me!
                    </mj-button>
                </mj-column>
            </mj-section>
        </mj-container>
    </mj-body>
</mjml>
```

Rendered with MJML 3.3.3:



Rendered with MJML 3.3.4:



A possible fix would be to add a space in front of the thin right arrow: `' ->'`

What do you think? Any chance we can get this fixed in 3.3.5? ;)

iRyusa:
Well nice catch, a 3.3.5 is totally needed here, if you want to do a PR on that so we can credit you ?

PR #778, opened 2017-08-14
Removes commas on multiple mjContent items

### LINKED ISSUE #675: Creating a component renders unwanted commas
Hi there, 

Thanks for the great work on the project, really like it. However when creating a custom component using the jsx syntax, we have some strange behaviour. 

The following is extracted from the `render` function : 
```
<Column width="50%">
    <Text font-size="14" align="center" padding="0" font-weight="700">
        test
        <a href="tel:0123456789" style={ {color: "#67cfd6"} }>0123456789</a>
        <span style={ {color:"#000000"} }>(appel gratuit)</span>
    </Text>
</Column>
```

This part gets rendered as followed : 
```
test,012345789,(appel gratuit)
```

It seems the mj-text component adds the commas when rendering each element.

etiennemarais:
## Summary

Fixes #675 

### Changes

- ~Added a check in the text component for children being an array~
- Reverted the text component only change in favour of the core content
- Added the join to the `mjContent` method to solve for any component
- Join the array of children on an empty separator instead of letting react's `__html` automatically decide that for us and add commas.

### Linting output success

```
╰─ yarn run lint
yarn run v0.27.5
$ eslint .
Done in 7.35s.
```

### DIFF
--- a/packages/mjml-core/src/decorators/MJMLElement.js
+++ b/packages/mjml-core/src/decorators/MJMLElement.js
@@ -78,12 +78,14 @@ function createComponent (ComposedComponent) {
         return trim(content)
       }
 
-      return React.Children.map(this.props.children, child => {
+      const contentItems = React.Children.map(this.props.children, child => {
         if (typeof child === 'string') {
           return child
         }
         return ReactDOMServer.renderToStaticMarkup(child)
       })
+
+      return (Array.isArray(contentItems)) ? contentItems.join("") : contentItems;
     }
 
     inheritedAttributes () {
--- a/packages/mjml-core/src/decorators/MJMLElement.js
+++ b/packages/mjml-core/src/decorators/MJMLElement.js
@@ -85,7 +85,7 @@ function createComponent (ComposedComponent) {
         return ReactDOMServer.renderToStaticMarkup(child)
       })
 
-      return (Array.isArray(contentItems)) ? contentItems.join("") : contentItems;
+      return (Array.isArray(contentItems)) ? contentItems.join("") : contentItems
     }
 
     inheritedAttributes () {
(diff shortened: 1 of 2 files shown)


iRyusa:
Hi @etiennemarais and thanks for this awesome PR !

Can you add this instead in the mjContent function ? I think it will fix this bug for every component instead of only one

etiennemarais:
Hi, Thank you for the kind words. Are you happy with adding it for all the components? I was only aware that the `mj-text` component were misbehaving?

etiennemarais:
If I make this change in the `MJMLElement.js` file then it would look something like this:

```

const contentItems = _react2.default.Children.map(_this.props.children, function (child) {
          if (typeof child === 'string') {
            return child;
          }
          return _server2.default.renderToStaticMarkup(child);
        });

return contentItems.join("");
```

Are you happy with something like that?

iRyusa:
Looks good now !

I'm trying to test this asap 👍

packages/mjml-core/src/decorators/MJMLElement.js
```diff
@@ -78,12 +78,14 @@ function createComponent (ComposedComponent) {
...
 
-      return React.Children.map(this.props.children, child => {
+      const contentItems = React.Children.map(this.props.children, child => {
         if (typeof child === 'string') {
           return child
         }
         return ReactDOMServer.renderToStaticMarkup(child)
       })
+
+      return (Array.isArray(contentItems)) ? contentItems.join("") : contentItems;
```
iRyusa: Can you remove the `;` ?

iRyusa:
Just tested out, after the little extra `;` PR looks OK to me !

etiennemarais:
Ah thx, My bad on the semi colon, So used to it coding in PHP the whole day. Thanks for helping to test. This will hopefully close a few issues for people :D

PR #779, opened 2017-08-14
Fix misprint in the button docs

zouxuoz:
Better docs 😉


iRyusa:
Thanks for finding this @zouxuoz 👍

PR #797, opened 2017-08-30
[Hero content] float property doesn't support center

nicholas-c:
https://developer.mozilla.org/en/docs/Web/CSS/float

As per MDN docs, the `float: center` doesn't do anything, it would be replaced with `float: none`, however `none` is the default value making it redundant.

Happy camping.


iRyusa:
Hi @nicholas-c and thanks for this PR

PR #805, opened 2017-09-05
fix(mj-button): make entire button area clickable

### LINKED ISSUE #359: Entire button as link
Hi Guys,

Firstly - thanks for the great work in building mjml, it makes my life so much easier!

I'm not sure if this is possible or if it's already planned.

Currently, using `<mj-button>` with `href` results in the text on the button being linked, but not the button itself.  For a large button with small text, this results in a large area that's not clickable.

Is it possible to change how the button links work so that the entire button is a link, rather than just the text?

Maistho:
resolves #359 

Can I get some help with setting up litmus tests for this? Have only tested in the clients I have at hand.

### DIFF
--- a/packages/mjml-button/src/index.js
+++ b/packages/mjml-button/src/index.js
@@ -70,11 +70,11 @@ class Button extends Component {
         cursor: 'auto',
         fontStyle: mjAttribute('font-style'),
         height: mjAttribute('height'),
-        padding: defaultUnit(mjAttribute('inner-padding'), "px"),
         textAlign: mjAttribute('text-align')
       },
       a: {
         background: mjAttribute('background-color'),
+        borderRadius: defaultUnit(mjAttribute('border-radius'), "px"),
         color: mjAttribute('color'),
         fontFamily: mjAttribute('font-family'),
         fontSize: defaultUnit(mjAttribute('font-size')),
@@ -83,6 +83,8 @@ class Button extends Component {
         lineHeight: mjAttribute('line-height'),
         textDecoration: mjAttribute('text-decoration'),
         textTransform: mjAttribute('text-transform'),
+        padding: defaultUnit(mjAttribute('inner-padding'), "px"),
+        display: "inline-block",
         margin: "0px"
       }
     })


packages/mjml-button/src/index.js
```diff
@@ -83,6 +83,8 @@ class Button extends Component {
         lineHeight: mjAttribute('line-height'),
         textDecoration: mjAttribute('text-decoration'),
         textTransform: mjAttribute('text-transform'),
+        padding: defaultUnit(mjAttribute('inner-padding'), "px"),
```
iRyusa: Well sadly, most of desktop outlook doesn't support padding on `a` element even with an `inline-block`

rogierslag:
This has been tried a number of times already.... https://github.com/mjmlio/mjml/search?q=button+clickable&type=Issues&utf8=%E2%9C%93

Maistho:
@rogierslag Yes it has, because it's an important feature. I'm confident that there is a way to solve this.

According to [this article from litmus](https://litmus.com/blog/a-guide-to-bulletproof-buttons-in-email-design), there are multiple approaches that work well for most clients.

iRyusa:
Well we're using the 2nd technique as described in this article, this is the "most" bulletproof buttons with the less constraints. Other can have some issue with multi-line buttons, or cannot be used inside a VML background

ngarnier:
Yes, and because the button will break in various scenarios (used over a background image or containing multiline content like @iRyusa said, which is pretty common), those approaches aren't acceptable. As MJML is a framework, it has to stay as standard as possible to support the various use-cases.

However, if you know that one of those drawbacks is ok for you (for example you never use buttons over background images), you can create your own custom component and we'll be happy to feature it on https://mjml.io/community. 

If you do find an approach that does work everywhere though, you're more than welcome to share it with us and we'll happily integrate it!

Maistho:
I've worked a bit more on it, and am considering different solutions. Multi line seems really difficult to get looking correct.

Personally I'm fine with the larger distance between lines on Outlook in order to improve the clickability of the buttons, but I can see why others wouldn't be.

I'll share litmus tests later today.

ngarnier:
Great news! Feel free to use the buttons in the styleguide for your tests so it's easy to compare the support before/after: https://github.com/mjmlio/mjml/blob/master/packages/mjml/styleguide.mjml#L187-L281

iRyusa:
Hi there, 

I'm closing this PR for now because we won't accept any PR (for non breaking bugs) on MJML3.
Feel free to reopen on the new `next` branch if you got some improvements for MJML4 !

PR #817, opened 2017-09-23
Prevent a problematic char from breaking HTML comments

witzig:
The thin right arrow (`->`) added to `DANGEROUS_CHARS` in PR #751 breaks the closing sequence (`-->`) of HTML comments, which in turn breaks the postrender functions and the resulting HTML.

This PR fixes the issue by adding a space in front of the thin right arrow: `' ->'`

- - -

``` xml
<mjml>
    <mj-body>
        <mj-container background-color="#eee">
            <mj-raw>
                <!-- some comment -->
            </mj-raw>
            <mj-section>
                <mj-column>
                    <mj-button >
                        Don't click me!
                    </mj-button>
                </mj-column>
            </mj-section>
        </mj-container>
    </mj-body>
</mjml>
```

Rendered with MJML 3.3.3:

Rendered with MJML 3.3.4:

Rendered with the patch:

fix

### DIFF
--- a/packages/mjml-core/src/MJMLRenderer.js
+++ b/packages/mjml-core/src/MJMLRenderer.js
@@ -23,7 +23,7 @@ import React from 'react'
 import ReactDOMServer from 'react-dom/server'
 import warning from 'warning'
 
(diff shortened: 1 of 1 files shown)


iRyusa:
Thanks you so much for this PR, I've just added the space on the opening tag too !

PR #824, opened 2017-10-01
Fix social align attribute

slavanga:


iRyusa:
Hi @slavanga 

Thanks for this PR, but MJML 3 won't receive fix anymore (only major breaking bug), as we want to focus on MJML4 release.

As far as I know, this align attribute on outlook conditionnal table doesn't do anything and should be removed, so this won't break anything if we don't merge this.

I apologize for closing this PR and hope it won't discourage you to contribute on this project again !

PR #825, opened 2017-10-03
Add background-position to mjml-section

GuillaumeBadi:


iRyusa:
Hi @GuillaumeBadi 

`background-position` isn't supported in Yahoo / Outlook.com & Desktop outlook, that would add some confusion about the support. 

So I think we won't accept this attribute, you can add it for you own need through mj-style and css-class.

Thanks for this PR even if we won't merge it.

PR #829, opened 2017-10-06
mjml-button padding fix

rafal-olszewski:
Moved padding from `td` element to `a` element, otherwise the clickable area is only the `a` element which is not the actuall button, but only the text.

After quick line switch, button should be rendered correctly.


iRyusa:
Hi @rafal-olszewski 

Thanks for this PR, but this has been discussed many times cf ( https://github.com/mjmlio/mjml/pull/805, https://github.com/mjmlio/mjml/pull/663, https://github.com/mjmlio/mjml/pull/557)

TLDR; Desktop outlook only accepts padding on TD, moving the padding on the A will make buttons either broken/ugly or unusable

Thanks a lot again, but I'm closing this PR

PR #830, opened 2017-10-08
Fix table rendering

mik-laj:


iRyusa:
Thanks for fixing the documentation on mj-social @mik-laj !

PR #834, opened 2017-10-10
Add colspan/rowspan to legacy attrs

mik-laj:


iRyusa:
Hi @mik-laj 

Well those attribute aren't used in the default components so I don't think I will merge that.
MJML3 dev has stopped, so we won't merge anything inside 3 branch (except for breaking bugs) to focus on MJML4 now.

Issue #846, opened 2017-10-20
mj-button as block


ogonkov:
Button without `display: block` on an anchor looks slightly out of centre (vertical).

Is there any special reason to not make link in button a block element? (i.e. some weird Outlook bug maybe?)

Link as block (on the left) vs default inline (on the right)
![anchor_as_block_vs_inline](https://user-images.githubusercontent.com/1091927/31813291-76290e8e-b58e-11e7-8c58-55a7955cc45c.png)

iRyusa:
Hi @ogonkov 

There's an open issue about this https://github.com/mjmlio/mjml/issues/359 I don't think adding display block on link would add something because the padding is done on the wrapping `td`

Full-width button has been discussed but there's no real bulletproof technique yet discovered to make full-width button.

For the vertical align it might add a bigger box than the line-height of the anchor that could be why you have some weird thing with vertical alignment

ogonkov:
@iRyusa It's a different issue. I'm about vertical alignment, that looks broken, if link is an inline element.

ogonkov:
More detailed illustration. Same button, the only difference is anchor `display` property.
![2017-10-20 12 22 09](https://user-images.githubusercontent.com/1091927/31814168-6baaa384-b591-11e7-9346-64863ddcc542.png)

iRyusa:
Well I'm unable to reproduce here https://mjml.io/try-it-live/HJ0fbSPaZ

ogonkov:
@iRyusa i use html beautify to avoid long lines (we have issues with it just 2 days ago), while default MJML output didn't have any spaces around link

This could be the reason that it's works in playground, and broken for our emails

ogonkov:
I will return with some code, couldn't reproduce it :/

ogonkov:
Localize it. The issue were with our web ui, it has 160% line height, that buttons `table` inherits, and broke vertical align.

PR #848, opened 2017-10-20
Ensure proper line-height of mj-button

ogonkov:
Parent line-heigh (from web mail client, for example) could visually broke mj-button vertical align

Demo from #846

https://mjml.io/try-it-live/rkMv8LvTb

### DIFF
--- a/packages/mjml-button/src/index.js
+++ b/packages/mjml-button/src/index.js
@@ -56,7 +56,6 @@ export default class MjButton extends BodyComponent {
       table: {
         'border-collapse': 'separate',
         width: this.getAttribute('width'),
-        // Ensure that wrong line-height wouldn't be inherited
         'line-height': '100%',
       },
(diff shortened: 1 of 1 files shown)


packages/mjml-button/src/index.js
```diff
@@ -56,6 +56,8 @@ export default class MjButton extends BodyComponent {
       table: {
         'border-collapse': 'separate',
         width: this.getAttribute('width'),
+        // Ensure that wrong line-height wouldn't be inherited
```
iRyusa: Can we remove the comment here ?

ogonkov: Probably, i'm personally prefer comments to git-blaming to trace the reason why code were added

I can remove it, if you want

ogonkov:
Fixed silly typo in first message

PR #854, opened 2017-10-24
Fix broken link to `try-it-live` in README

nelson6e65:
In the README file, the link to https://mjml.io/try-it-live is broken; points to an relative path instead.


iRyusa:
Nice catch thanks for this pr !

Issue #660, opened 2017-05-02
mjml --validate does not parse .mjmlconfig


sastraxi:
mjml version: 3.1.1
my .mjmlconfig:
```
{
  "packages": [
    "./lib/mjml/TrackingPixel.js"
  ]
}
```
my custom component (before babel):
```
import { MJMLElement } from 'mjml-core';
import React from 'react';
import Image from 'mjml-image';

const Component = ({ mjAttribute }) => // eslint-disable-line
  <Image
    src={mjAttribute('url')}
    width={1}
    height={1}
  />;

const TrackingPixel = MJMLElement(Component);

TrackingPixel.tagName = 'tracking-pixel';
TrackingPixel.parentTag = ['mj-container'];
TrackingPixel.endingTag = false;
TrackingPixel.defaultMJMLDefinition = {
  content: '',
  attributes: {
    url: '',
  },
};

export default TrackingPixel;
```
an example template:
```
<mjml>
  <mj-body>
    <mj-container>
      <tracking-pixel url={trackingurl} />
      <mj-section>
        <mj-column>

          <mj-image width="100" src="..."></mj-image>

          <mj-divider border-color="#F45E43"></mj-divider>

          <mj-text font-size="20px" color="#F45E43" font-family="helvetica">Password Reset</mj-text>
          
          <mj-text>
            A little bird told us that you're unable to log in. We're here to help!
          </mj-text>
          
          <mj-button href="{url}" font-family="Helvetica" background-color="#f45e43" color="white">
            Reset my password
          </mj-button>

        </mj-column>
      </mj-section>
    </mj-container>
  </mj-body>
</mjml>
```
Validating the template before compiling it to HTML gives me this:
```
cameron@hans $ mjml --validate templates/email/reset-password.mjml 
Line 4 (tracking-pixel) — Element tracking-pixel doesn't exist or is not registered
```
But compiling it to HTML works, and the tracking pixel shows up in the compiled template.
```
cameron@hans $ mjml -r ./templates/email/reset-password.mjml && cat reset-password.html
<!doctype html>
<html xmlns="http://www.w3.org/1999/xhtml" xmlns:v="urn:schemas-microsoft-com:vml" xmlns:o="urn:schemas-microsoft-com:office:office">
<head>
  <title></title>
  <!--[if !mso]><!-- -->
  <meta http-equiv="X-UA-Compatible" content="IE=edge">
  <!--<![endif]-->
[...]

PR #855, opened 2017-10-25
check .mjmlconfig for custom components to validate against

dhcdata:
The execution path divergence between validate and render happens in the `bin/mjml` file.  Ideally this would run before so that it doesn't have to be called in each execution path. Because of this, the PR attached is not the most elegant way of solving issue #660 , but it works.


dhcdata:
I see from #804 that `validate` is being removed.

iRyusa:
Hi @dhcdata 

`mjmlconfig` and `validate` aren't yet available in next MJML4 build yes ! But we should port it back soon (if you want to get involved into it would be great ! feel free to DM me in Slack 👍) 

We're only merging breaking bugs in MJML3 branch now, so i'm gonna close this PR but will take it back into `next` when we're porting both feature back into the branch, you'll be credit for this ofc !

PR #865, opened 2017-11-01
Allow yargs configuration to be composable

dhcdata:
By importing the yargs object from a separate file, a developer can easily extend mjml for their own use.

Since `node_modules/.bin/mjml` is just running several scripts, it's just as easy for a developer to write their own shell script as such:

```
#!/usr/bin/env node

require('mjml')         // register all of mjml base components
require('./lib/index')  // register custom components
require('./lib/client') // register custom command line options
require('mjml-cli')     // run program

```

### DIFF
--- a/packages/mjml-button/src/index.js
+++ b/packages/mjml-button/src/index.js
@@ -91,31 +91,34 @@ export default class MjButton extends BodyComponent {
     return `
       <table
         ${this.htmlAttributes({
-    align: this.getAttribute('align'),
-    border: '0',
-    cellpadding: '0',
-    cellspacing: '0',
-    role: 'presentation',
-    style: 'table',
-  })}
+          align: this.getAttribute('align'),
+          border: '0',
+          cellpadding: '0',
+          cellspacing: '0',
+          role: 'presentation',
+          style: 'table',
+        })}
       >
         <tr>
           <td
             ${this.htmlAttributes({
-    align: 'center',
-    bgcolor: this.getAttribute('background-color') === 'none' ? undefined : this.getAttribute('background-color'),
-    role: 'presentation',
-    style: 'td',
-    valign: this.getAttribute('vertical-align'),
-  })}
+              align: 'center',
+              bgcolor:
+                this.getAttribute('background-color') === 'none'
+                  ? undefined
+                  : this.getAttribute('background-color'),
+              role: 'presentation',
+              style: 'td',
+              valign: this.getAttribute('vertical-align'),
+            })}
           >
             <${tag}
               ${this.htmlAttributes({
-    href: this.getAttribute('href'),
-    rel: this.getAttribute('rel'),
-    style: 'content',
-    target: tag === 'a' ? '_blank' : undefined,
-  })}
+                href: this.getAttribute('href'),
+                rel: this.getAttribute('rel'),
+                style: 'content',
+                target: tag === 'a' ? '_blank' : undefined,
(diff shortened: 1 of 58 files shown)


packages/mjml-button/src/index.js
```diff
@@ -56,7 +56,6 @@ export default class MjButton extends BodyComponent {
       table: {
         'border-collapse': 'separate',
         width: this.getAttribute('width'),
-        'line-height': '100%',
```
iRyusa: That shouldn't be removed I think

iRyusa:
Hi @dhcdata ( much thanks for you contribution lately ! )

Can you take a look at `next-mj-hero` branch, the cli becomes a function, so what you can do is add an `yargs` instance as an argument and will default to the current one. I think this would be a good compromise.

On a feature side, is this really something we want ? I don't think other project has such thing to offer (meaning: override the cli behavior). 

As you describe I don't really see the need to override the yargs instance to extend it, what do you think you would add and why you can't just parse your arguments with your own `yags` instance in your script ? Does it makes sense to you, or maybe i'm completely wrong about the use case

dhcdata:
Can you point to the file in `next-mj-hero` where cli is a function?  I'm not seeing a difference off hand.

As far as the feature, perhaps describing the use case that brought me to mjml might help.

Our company is using a CRM that has a horrible template gui for their email product. In one mode, I need to be able to use watch to output the entire html so that the developer can get the instant feedback mjml provides. In the second mode, I need to output only the non-template html to paste into the CRM vendor's WYSIWYG editor.

I don't technically need yargs to be composable to accomplish this. I do need it in order to list the custom option in the `--help` output

iRyusa:
So here it is : https://github.com/mjmlio/mjml/commit/e29d085c70d10018f27af2a0bcea339f59b0e3fb#diff-09f789cd5e5be6dbabfa50f310428f27R16

> The passed argument changes the behavior of the component rendering.

That sounds way more as a fork of mjml-cli to me rather than fitting into common use case ?

dhcdata:
It changes the rendering behavior of my custom component.

iRyusa:
It's weird that the cli will have some controls over the `core` behavior. Did you tried to achieve something similar with ENV variables inside your custom component ?

`MY_SPECIAL_COMPONENT=my_mode`
then
`if(process.env.MY_SPECIAL_COMPONENT == 'my_mode) { ... }`

It would totally bypass the cli override only for a single option ?

dhcdata:
Like I said, I can do `process.argv` to get access to the flag I need. It just won't allow my colleagues who are not tech savvy to see the option when they run `./mjml --help`

iRyusa:
Well, as I said : 

> On a feature side, is this really something we want ? I don't think other project has such thing to offer (meaning: override the cli behavior).

I think, we're in a case where you want to fork the cli and maintain your own fork for this specific need.
Composable cli would be a pain to maintain and to offer enough flexibility to keep everyone happy.

I'm not a big fan of it, but passing yargs instance as an option of the default func of `next-mj-hero` (https://github.com/mjmlio/mjml/commit/e29d085c70d10018f27af2a0bcea339f59b0e3fb#diff-09f789cd5e5be6dbabfa50f310428f27R16)  would be the only way to do this without too much overhead

dhcdata:
This PR is not necessary.  `yargs` is the composable object. Chaining options onto that in the consumer's project results in the proper help file and parse.

PR #869, opened 2017-11-03
Enhancement — Allow mj-image to take string values

mdpwilliams:
This enhancement will allow `mjml-image` to put a string directly into the markup.

The goal is to allow:

```
<mj-image
  src="{{image.src}}"
  width="{{image.width}}" />
```

To produce markup with `{{image.width}}` within it, so that it could be replaced by another templating language like Handlebars or MJML.


iRyusa:
Hi @mdpwilliams 
Can you reopen this on `next` branch, 3.3.X will only be update for major breaking bug.

About this feature, I'm not a big fan of adding those kind of check only to comply for a templating language, but I think we could accept that.

PR #870, opened 2017-11-05
allow skeleton to be overridden

dhcdata:
This PR allows the template literal to be overridden by the client's application

### DIFF
--- a/packages/mjml-core/src/helpers/skeleton.js
+++ b/packages/mjml-core/src/helpers/skeleton.js
@@ -2,67 +2,71 @@ import buildPreview from './preview'
 import { buildFontsTags } from './fonts'
 import { buildMediaQueriesTags } from './mediaQueries'
 
-export default function skeleton(options) {
-  const {
-    breakpoint = '480px',
-    content = '',
-    fonts = {},
-    mediaQueries = {},
-    preview,
-    title = '',
-    style,
-  } = options
+export let template = ({
+  breakpoint = '480px',
+  content = '',
+  fonts = {},
+  mediaQueries = {},
+  preview,
+  title = 'bob',
+  style,
+}) =>`
+<!doctype html>
+<html xmlns="http://www.w3.org/1999/xhtml" xmlns:v="urn:schemas-microsoft-com:vml" xmlns:o="urn:schemas-microsoft-com:office:office">
+  <head>
+    <title>
+      ${title}
+    </title>
+    <!--[if !mso]><!-- -->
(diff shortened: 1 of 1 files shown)


iRyusa:
I would prefer to have an extra option on `mjml-core`/ `mjml2html` option to override the skeleton instead of creating multiple set/get

dhcdata:
I'm not sure how to implement what you're suggesting without a setter somewhere given template literals are frozen.  If you can explain in pseudo code, I'd be happy to take a stab at implementing.

iRyusa:
No problem !

Here you have engine option : https://github.com/mjmlio/mjml/blob/next/packages/mjml-core/src/index.js#L40 

So you have something like `skeleton: defaultSkeleton` that will be a function (taking content, and global data), and here https://github.com/mjmlio/mjml/blob/next/packages/mjml-core/src/index.js#L204 you will call this skeleton from options

dhcdata:
With the current setup, the only way to utilize that would be to write the function on the command line.

To fix that, [`DEFAULT_OPTIONS`](https://github.com/mjmlio/mjml/blob/eebbbbe3f274f430535af7d1082570ce483e6a03/packages/mjml-cli/src/client.js#L16) would need to be imported as an object from another file.

Is that satisfactory? That should allow me to simply override with my own default option.

iRyusa:
> To fix that, DEFAULT_OPTIONS would need to be imported as an object from another file. 
Cli options aren't tied to engine options ? I think default options are default cli options, at some point not "everything" should be as customizable.

Skeleton could be either be a function or a path to the file that will export the skeleton template ?

dhcdata:
Perhaps there is misunderstanding.

```
mjml-cli
└───src
   │   client.js // imports from ./helper/defaultOptions
   └───helpers
       │   defaultOptions.js
       │   fileContext.js
```

iRyusa:
Well same issue with the other PR, I think default option override is such an edge case that I don't really understand the need of letting someone override it. `cli` is meant to only allow basic I/O things and not aim to be a massive configurable/extensible thing. 

It has an interface to modify the option you'd want to, so why not using it ?

dhcdata:
Where is there an interface for me to override the skeleton? There is in the 3.x branch.

iRyusa:
Not really, overriding internal stuff in general isn't really a good practice. 

As I proposed : 

https://github.com/mjmlio/mjml/pull/870#issuecomment-342105742 + 

> Skeleton could be either be a function or a path to the file that will export the skeleton template ?

dhcdata:
Thanks for the direction. I believe I implemented as suggested.

PR #872, opened 2017-11-06
allow config option to accept a JSON string

### LINKED ISSUE #871: Yargs does not accept type: 'object'
[`mjml-cli/client.js L57-61`](https://github.com/mjmlio/mjml/blob/eebbbbe3f274f430535af7d1082570ce483e6a03/packages/mjml-cli/src/client.js#L57)
```
    c: {
      alias: 'config',
      type: 'object',
      describe: 'Option to pass to mjml-core',
    },
```

type: 'object' is not a valid [`yargs option type`](http://yargs.js.org/docs/#api-optionkey-opt). Must be one of:
```
'array': synonymous for array: true, see array()
'boolean': synonymous for boolean: true, see boolean()
'count': synonymous for count: true, see count()
'number': synonymous for number: true, see number()
'string': synonymous for string: true, see string()
```

Probably need to utilize the [`coerce`](http://yargs.js.org/docs/#api-coercekey-fn) API function. I don't have enough experience with `yargs` to be sure this is the easiest approach.

dhcdata:
Fixes #871 . To use on the command line, enclose the json object in single quotes like so: `-c '{ "beautify": false}'`

### DIFF
--- a/packages/mjml-cli/src/client.js
+++ b/packages/mjml-cli/src/client.js
@@ -56,7 +56,8 @@ const argv = yargs
     },
     c: {
       alias: 'config',
-      type: 'object',
+      coerce: JSON.parse,
+      type: 'string',
       describe: 'Option to pass to mjml-core',
     },
     version: {


iRyusa:
HI @dhcdata 

Can we update this to remove the type & coerce as we mentioned in the original issue to avoid an issue when they gonna add the `object` type support ? thanks !

dhcdata:
Why do you suspect they're going to add `object` type support any time soon? yargs/yargs#890 was opened by the owner of the repo with a tag that he hopes someone else will implement.

I suspect your reasoning is because the coerce function in this PR probably overrides the dot-notation syntax. An improved coerce function should resolve it so that it works as expected under both.  I'm happy to remove it since #873  will provide an alternative option for a nested config object.

iRyusa:
>  Update? That's the entire PR. You mean close?

Well, It was only to credit you that `object` isn't a valid option, up to you.

> Why do you suspect they're going to add object type support any time soon? yargs/yargs#890 was opened by the owner of the repo in June with a tag that he hopes someone else will implement.

Better safe than sorry, always avoid possible side effect :)

> I suspect your reasoning is because the coerce function in this PR probably overrides the dot-notation syntax.

I think the dot syntax is the best approach to allow editing the config object in the cli, so yes. 
But sending JSON as plain text of a cli isn't a really user friendly, and easy to edit in a CMD line.
I feel like we should be careful when providing many way to input options, and would bloat the `cli`

PR #873, opened 2017-11-06
implement skeleton override as suggested in #870

dhcdata:
> Skeleton could be either be a function or a path to the file that will export the skeleton template ?

implement skeleton override as suggested in #870 . This implements the former.


packages/mjml-core/src/index.js
```diff
@@ -38,6 +38,7 @@ export default function mjml2html(mjml, options = {}) {
     },
     keepComments,
     minify = false,
+    skeleton = (opt) => defaultSkeleton(opt),
```
iRyusa: defaultSkeleton should be enough here

packages/mjml-core/src/index.js
```diff
@@ -38,7 +42,7 @@ export default function mjml2html(mjml, options = {}) {
     },
     keepComments,
     minify = false,
-    skeleton = (opt) => defaultSkeleton(opt),
+    skeleton = (opt) => defaultSkeleton,
```
iRyusa: `skeleton = defaultSkeleton`

packages/mjml-core/src/index.js
```diff
@@ -25,6 +25,10 @@ export default function mjml2html(mjml, options = {}) {
   let content = ''
   let errors = []
 
+  if(typeof options.skeleton === 'string') {
+    options.skeleton = (opt) => require(options.skeleton)
```
iRyusa: `options.skeleton = require(options.skeleton)`

packages/mjml-core/src/index.js
```diff
@@ -26,7 +26,7 @@ export default function mjml2html(mjml, options = {}) {
   let errors = []
 
   if(typeof options.skeleton === 'string') {
-    options.skeleton = (opt) => require(options.skeleton)
+    options.skeleton = require(options.skeleton)
```
iRyusa: Mhh, only issue is the require will be relative to the file calling it, might need to pass the process.cwd if it's a relative path ?

PR #884, opened 2017-11-14
Fix bullets and add code formatting

asmockler:
I noticed a couple problems and inconsistencies with the validator documentation - this PR fixes those. ✏️ 

I'm not sure if this PR should be against `next` or `master` and couldn't find any info. I opened it against `master` since this is a change to the current docs - let me know if I should close and open it against `next` instead. Thanks 😄


iRyusa:
hi @asmockler 

Sorry about this, i should change master to current next, because master is now legacy 3.3.x branch.

Can you cherry pick it into next ? So we can merge it thanks !

PR #912, opened 2017-12-08
MJML4 : Navbar

kmcb777:


packages/mjml-core/src/createComponent.js
```diff
@@ -105,7 +105,7 @@ export class BodyComponent extends Component {
   }
 
   styles(styles) {
-    let stylesObject = this.getStyles
+    let stylesObject = this.getStyles()
```
iRyusa: Mh, empty Object here or just keep the `let stylesObject` alone

packages/mjml-core/src/helpers/conditionalTag.js
```diff
@@ -10,3 +12,11 @@ export default function conditionalTag(content, negation = false) {
     ${negation ? endNegationConditionalTag : endConditionalTag}
   `
 }
+
+export const msoConditionalTag = (content, negation = false) => {
```
iRyusa: I think exporting another `function` would be better for clarity 👍

packages/mjml-core/src/index.js
```diff
@@ -105,14 +106,18 @@ export default function mjml2html(mjml, options = {}) {
...
     const component = initComponent({
-      name: node.tagName,
+      name,
       initialDatas: {
         ...parseMJML(node),
         context,
       },
     })
 
+
```
iRyusa: too much empty line :)

packages/mjml-core/src/index.js
```diff
@@ -167,6 +172,9 @@ export default function mjml2html(mjml, options = {}) {
         className
       ] = `{ width:${parsedWidth}${unit} !important; }`
     },
+    addHeadSyle(componentName, headStyle) {
```
iRyusa: I wonder if with an arrow function you could omit the componentName here ?

packages/mjml-navbar/src/Navbar.js
```diff
@@ -0,0 +1,181 @@
...
+      .mj-menu-checkbox[type="checkbox"] ~ .mj-inline-links > a { display:block!important; }
+      .mj-menu-checkbox[type="checkbox"]:checked ~ .mj-menu-trigger .mj-menu-icon-close { display:block!important; }
+      .mj-menu-checkbox[type="checkbox"]:checked ~ .mj-menu-trigger .mj-menu-icon-open { display:none!important; }
+    }
+  `
+
+  getStyles () {
+    const { mjAttribute, defaultUnit, getPadding } = this.props
+
+    return merge({}, baseStyles, {
```
iRyusa: Can you now directly return the object instead of merging like here : https://github.com/mjmlio/mjml/blob/next/packages/mjml-social/src/SocialElement.js#L84-L113

packages/mjml-navbar/src/Navbar.js
```diff
@@ -0,0 +1,181 @@
...
+      .mj-menu-checkbox[type="checkbox"]:checked ~ .mj-inline-links,
+      .mj-menu-checkbox[type="checkbox"] ~ .mj-menu-trigger { display:block!important; max-width:none!important; max-height:none!important; font-size:inherit!important; }
+      .mj-menu-checkbox[type="checkbox"] ~ .mj-inline-links > a { display:block!important; }
+      .mj-menu-checkbox[type="checkbox"]:checked ~ .mj-menu-trigger .mj-menu-icon-close { display:block!important; }
+      .mj-menu-checkbox[type="checkbox"]:checked ~ .mj-menu-trigger .mj-menu-icon-open { display:none!important; }
+    }
+  `
+
+  getStyles () {
+    const { mjAttribute, defaultUnit, getPadding } = this.props
```
iRyusa: getPadding/defaultUnit aren't used here I think ?

packages/mjml-navbar/src/Navbar.js
```diff
@@ -0,0 +1,181 @@
...
+          </span>
+        </label>
+      </div>
+    `
+  }
+
+
+  render () {
+    return `
+      <div>
```
iRyusa: I wonder if this div was only needed for `react` purpose can you check ?

packages/mjml-navbar/src/Navbar.js
```diff
@@ -0,0 +1,181 @@
...
+          : '' }
+        ${conditionalTag(`
+          <table role="presentation" border="0" cellpadding="0" cellspacing="0" align="${this.getAttribute('align')}">
+            <tr>
+        `)}
+        <div
+          ${this.htmlAttributes({
+            class: "mj-inline-links",
+            style: this.htmlAttributes('div'),
+            'data-align': this.getAttribute('align')
```
iRyusa: I don't think `data-align` is needed anymore

packages/mjml-navbar/src/NavbarLink.js
```diff
@@ -0,0 +1,101 @@
+import { BodyComponent } from 'mjml-core'
+import url from 'url'
+import { merge } from 'lodash'
+
+import conditionalTag from 'mjml-core/lib/helpers/conditionalTag'
+
+const baseStyles = {
```
iRyusa: Same here

packages/mjml-navbar/src/Navbar.js
```diff
@@ -0,0 +1,181 @@
...
+
+  render () {
+    return `
+      <div>
+        ${ this.getAttribute('hamburger') === 'hamburger'
+          ? this.renderHamburger()
+          : '' }
+        ${conditionalTag(`
+          <table role="presentation" border="0" cellpadding="0" cellspacing="0" align="${this.getAttribute('align')}">
+            <tr>
```
iRyusa: I think outlook will have an issue here div > table > tr > div, missing a `td` ?

packages/mjml-navbar/src/NavbarLink.js
```diff
@@ -0,0 +1,101 @@
...
+    `
+  }
+
+  render() {
+    const padding = this.getAttribute('padding')
+    const height = this.getAttribute('height')
+
+    return height
+      ? `
+        ${conditionalTag(`
```
iRyusa: same here  with outlook tags, are you 100% sure that we handle that case in then navbar parent ?

packages/mjml-core/src/createComponent.js
```diff
@@ -105,7 +105,7 @@ export class BodyComponent extends Component {
   }
 
   styles(styles) {
-    let stylesObject = this.getStyles
+    let stylesObject = null
```
iRyusa: Just `let stylesObject` is enough :p

packages/mjml-core/src/index.js
iRyusa: Well I think we can keep it, might rename it as `identifier` instead  of `componentName`

packages/mjml-navbar/src/NavbarLink.js
```diff
@@ -0,0 +1,92 @@
...
+      </a>
+    `
+  }
+
+  render() {
+    const padding = this.getAttribute('padding')
+
+    return `
+        ${conditionalTag(`
+          <td style="padding: ${padding}">
```
iRyusa: You should use some styling from getStyles object with `this.htmlAttributes` here, ( plus padding-top, bottom, ...)  are missing 👍

PR #923, opened 2017-12-18
Mjml accordion

kmcb777:


packages/mjml-accordion/src/Accordion.js
```diff
@@ -0,0 +1,100 @@
+import { BodyComponent } from 'mjml-core'
+
+// const parentTag = ['mj-column', 'mj-hero-content']
```
iRyusa: To remove

packages/mjml-accordion/src/AccordionElement.js
```diff
@@ -0,0 +1,110 @@
...
+      'icon-unwrapped-alt',
+    ].reduce((res, val) => ({
+      ...res,
+      [val]: this.getAttribute(val)
+    }), {})
+
+    const result = []
+
+    if (!find(children, {tagName: 'mj-accordion-title'})) {
+      result.push((new AccordionTitle({ attributes: childrenAttr })).render())
```
iRyusa: Might need to append the context in it  ?

PR #938, opened 2018-01-09
Mjml carousel

kmcb777:


packages/mjml-carousel/src/Carousel.js
```diff
@@ -0,0 +1,326 @@
...
+    if (!length) return ''
+
+    const carouselCss = `
+    .mj-carousel {
+      -webkit-user-select: none;
+      -moz-user-select: none;
+      user-select: none;
+    }
+
+    .mj-carousel-icons-cell {
```
iRyusa: You don't need to prefix it with the carouselID too ?

kmcb777: No, this class is not involved with the checked/hover behaviors so it's fine

iRyusa: Ok but cell width is related to current carousel no (cf the getAttribute right after) ?

kmcb777: That's actually very true, i'll fix that

PR #939, opened 2018-01-09
allow multiple instances of a component to add their own style

kmcb777:
Cleaner way to allow each Carousel to have its own head style.
Also includes the README for navbar, carousel and accordion. Only the navbar one needed a slight modification since tagName has changed, otherwise all api remains the same as v3.


packages/mjml-core/src/components.js
```diff
@@ -13,7 +13,10 @@ export function initComponent({ initialDatas, name }) {
     const component = new Component(initialDatas)
 
     if (component.headStyle) {
-      component.context.addHeadSyle(name, component.headStyle())
+      component.context.addHeadSyle(name, component.headStyle.bind(component))
```
iRyusa: Arrow function like this don't work ?

```js
class Component {
  headStyle = () => {
    // ...
  }
}
```

to prevent the bind ?

PR #961, opened 2018-01-22
Set minifyCSS to false

Michael-wm:
html-minifier is converting every color name to a hex-value. This causes font names like "Arial Black" to be turned into "Arial #000". Disabling minifyCSS prevents this.


iRyusa:
Sorry for the wait thanks for this PR @Michael-wm !

PR #1005, opened 2018-02-23
allow to cumulate css-classes from mj-classes - resolve #722

### LINKED ISSUE #722: Allow css-class to add multiple class names via multiple mj-class
Currently if you have `css-class` on a`mj-class` and use multiple `mj-class` on a component, only the last `css-class` is actually added to the HTML.

I think it would be good if it would add all the classes.

Example:
Currently- 
```
<mj-class name="foo" css-class="foo"..."
 <mj-class name="bar" css-class="bar"..."


<mj-text mj-class="foo bar"...
```

will only add the `bar` class to the HTML.

`<tr class="bar">`

Ideally it should add both classes:

`<tr class="foo bar">`

Thanks

kmcb777:


packages/mjml-core/src/index.js
```diff
@@ -136,12 +136,24 @@ export default function mjml2html(mjml, options = {}) {
...
       const attributesClasses = reduce(
         classes,
-        (acc, value) => ({
-          ...acc,
-          ...globalDatas.classes[value],
-        }),
+        (acc, value) => {
+          const mjClassValues = globalDatas.classes[value]
+          let multipleClasses = {}
+          if (acc['css-class'] && mjClassValues && mjClassValues['css-class']) {
```
iRyusa: Can you use lodash get here ?

Issue #1004, opened 2018-02-22
srcset is invalid


sallf:
Trying to use `srcset` in `<mj-image>` and am getting a 'srcset is invalid' error when I go to compile. Looks like it's no longer in `allowedAttributes`. Was this removed on purpose? Code below for reference:
```
<mj-column>
  <mj-image
    padding="0"
    src="image1.jpg"
    srcset="image1.jpg 900w, image2.jpg 1200w"
    alt=""
    href="https://example.com"
  />
</mj-column>
```
Thanks!

kmcb777:
The fix has been merge we can close this one 👍

PR #1015, opened 2018-03-01
Add srcset attribute to mj-image*

jtassin:
According to the documentation mj-image should be able to handle srcset attribute. It was in v3 but it isn't anymore the case in v4.

srcset attribute is back in this PR.


iRyusa:
Thanks for this @jtassin ! Fix #1004

PR #1026, opened 2018-03-07
Update README.md

oliviertassinari:
This is an undocumented breaking change: https://github.com/mjmlio/mjml/blob/master/packages/mjml/src/index.js#L44


ngarnier:
Thanks!

PR #1028, opened 2018-03-07
Remove doc for attribute `inner-padding`.

twisty:
Removed the `inner-padding` attribute from the doc for `mj-social-element`, because it isn't valid.


iRyusa:
Thanks @twisty for this !

PR #1030, opened 2018-03-08
mj-social: remove `icon-color` attrib from code and doc.

twisty:
Setting `icon-color` on a `<mj-social-element />` has no effect. Removed from doc and from the allowed attributes list. The correct way to set a colour of the icon is via the `background-color` attribute, which is already documented.


iRyusa:
Hi ! 

Can you update the doc with background-color instead of removing the attributes please ?

twisty:
Done... 😁

iRyusa:
👍 for the commit name 🐶

PR #1033, opened 2018-03-08
fix #1027 allow traverseMjml to ignore part of the tree

### LINKED ISSUE #1027: MJML 4.0 — Setting mj-body attributes in mj-head renders an empty div
Here's the code:
```
<mj-head>
    <mj-title>Title</mj-title>
    <mj-attributes>
        <mj-body background-color="#ffffff" />
    </mj-attributes>
 </mj-head>
```

Tested in the MJML macOS app and with the `mjml-core` package.

The error occurs the minute you put `<mj-body />` in `<mj-attributes />`. No need to add any attributes to the tag.

kmcb777:


packages/mjml-core/src/helpers/traverseMJML.js
```diff
@@ -3,9 +3,13 @@ import head from 'lodash/head'
 import map from 'lodash/map'
 import noop from 'lodash/noop'
 
-export default function traverseMJML(mjml, predicate = noop) {
+export default function traverseMJML(mjml, predicate = noop, ignoreTagContent = null) {
```
iRyusa: Would prefer to have an `options = {}` here instead of a named string.
String parameters can be hard to read when you don't have any context

packages/mjml-core/src/index.js
```diff
@@ -104,7 +104,7 @@ export default function mjml2html(mjml, options = {}) {
       break
   }
 
-  const mjBody = traverseMJML(mjml, child => child.tagName === 'mj-body')
+  const mjBody = traverseMJML(mjml, child => child.tagName === 'mj-body', 'mj-attributes')
```
iRyusa: Like here

iRyusa: When thinking twice as this, I think it would add an hard dep on 'mj-attributes'  in the core. I think you should add a `deepSearch` parameters that would only search in root level childrens

Issue #1048, opened 2018-03-14
Version 4: mj-image not supporting alt attribute


brigitschroeder:
After migrating to version 4 of mjml, the alt attributes of the mj-image tag are no longer appearing in the generated html. Also, if there is an href tag on the mj-image, the value of the href gets assigned to the alt tag.

**Reproduction Steps:**

1. Install latest mjml in directory
$ npm init -y && npm install mjml

2. Simple example of input.mjml

```
<mjml>
  <mj-head>
    <mj-title>Example</mj-title>
  </mj-head>
  <mj-body>
    <mj-section>
      <mj-column>
        <mj-image src="https://promo.spacefoundation.org/stardays/march2018/family-star-party.png" alt="This is my alt text"></mj-image>
        <mj-text>The alt text in the image above is missing.</mj-text>
      </mj-column>
    </mj-section>

    <mj-section>
      <mj-column>
        <mj-image src="https://promo.spacefoundation.org/stardays/march2018/family-star-party.png" alt="This is my alt text" href="https://www.discoverspace.org/events/family-star-party"></mj-image>
        <mj-text>The alt text in the image above has been replaced with href text.</mj-text>
      </mj-column>
    </mj-section>
  </mj-body>
</mjml>

```

3. Generate the html
$ mjml --watch input.mjml --output index.html

4. View the index.html file

**Expected behavior:**

Expected behavior is that the alt text is added to the img tag in the html.

**Observed behavior:**

The observed behavior is that the alt text is ignored in the first instance, and the href text is captured as the alt text in the second instance. 

```
<img height="auto" src="https://promo.spacefoundation.org/stardays/march2018/family-star-party.png" style="border:0;display:block;outline:none;text-decoration:none;width:100%;" width="550" />
```

```
<img alt="https://www.discoverspace.org/events/family-star-party" height="auto" src="https://promo.spacefoundation.org/stardays/march2018/family-star-party.png" style="border:0;display:block;outline:none;text-decoration:none;width:100%;" width="550" />
```

**MJML version:**

mjml-core: 4.0.0
mjml-cli: 4.0.0

**Email clients the bug is seen on:**

Tested in browser.

PR #1050, opened 2018-03-15
Fix alt attribute missing from image

garmeeh:
Previous to this code fix adding a `href` attribute to `mj-image` would render the alt tag. 

Now adding `alt` to `mj-image` will render the alt tag as expected.

Fixes: https://github.com/mjmlio/mjml/issues/1048


iRyusa:
Ew how did we miss :( that thanks a lot for the fix @garmeeh !

PR #1060, opened 2018-03-16
Try to fix as many eslint errors as possible

thomaschaaf:
These are left:

```/Users/thomaschaaf/develop/github/mjml/packages/mjml-core/src/helpers/jsonToXML.js
  4:22  error  'jsonToXML' is not defined  no-undef

/Users/thomaschaaf/develop/github/mjml/packages/mjml-migrate/src/migrate.js
   33:3  error  The body of a for-in should be wrapped in an if statement to filter unwanted properties from the prototype                                                             guard-for-in
   33:3  error  for..in loops iterate over the entire prototype chain, which is virtually never what you want. Use Object.{keys,values,entries}, and iterate over the resulting array  no-restricted-syntax
   55:3  error  The body of a for-in should be wrapped in an if statement to filter unwanted properties from the prototype                                                             guard-for-in
   55:3  error  for..in loops iterate over the entire prototype chain, which is virtually never what you want. Use Object.{keys,values,entries}, and iterate over the resulting array  no-restricted-syntax
   64:5  error  for..in loops iterate over the entire prototype chain, which is virtually never what you want. Use Object.{keys,values,entries}, and iterate over the resulting array  no-restricted-syntax
   76:3  error  for..in loops iterate over the entire prototype chain, which is virtually never what you want. Use Object.{keys,values,entries}, and iterate over the resulting array  no-restricted-syntax
   94:3  error  The body of a for-in should be wrapped in an if statement to filter unwanted properties from the prototype                                                             guard-for-in
   94:3  error  for..in loops iterate over the entire prototype chain, which is virtually never what you want. Use Object.{keys,values,entries}, and iterate over the resulting array  no-restricted-syntax
  113:3  error  for..in loops iterate over the entire prototype chain, which is virtually never what you want. Use Object.{keys,values,entries}, and iterate over the resulting array  no-restricted-syntax```

The jsonToXML.js is incorrect.. and will not work currently I don't understand the intention so I can't fix it. You should also rewrite the code so that it doesn't use for..in loops anymore :)


iRyusa:
Hi @thomaschaaf 

I will check that asap, thanks for the pr

kmcb777:
Hi @thomaschaaf thanks for your PR,
I just checked it quickly and the only problem i see is the functions that become `static` (getStyles, renderAfter). These have to remain instance methods because they are common to all components and used when processing, and on most components they need to access `this`.

iRyusa:
Well with no update on this one, we can close in favor of #1134 I think ?

Thanks for this contribution anyway

PR #1058, opened 2018-03-16
Make section alignable

thomaschaaf:
Currently all emails are centered. I would like to send emails which are not centered.

I have attached screenshots for this mjml code.
```
<mjml>
  <mj-body>
    <mj-section align="left">
      <mj-column>
        <mj-divider border-color="#F45E43"></mj-divider>
        <mj-text font-size="20px" color="#F45E43" font-family="helvetica">Hello World</mj-text>
      </mj-column>
    </mj-section>
  </mj-body>
</mjml>
```

PR #1066, opened 2018-03-21
Make section alignable

thomaschaaf:
In our A/B testing we have discovered that sending emails which look like they come from people "plain text like" emails have a better reply/interaction rate than emails which look like newsletters.
One factor that makes an email look like a mass-mailing/newsletter is that it is centered. We would like to align the block to the left.

For some reason the previous pr was closed: https://github.com/mjmlio/mjml/pull/1058


ngarnier:
Thanks for explaining the use case @thomaschaaf. However, as I showed in the initial PR, those changes introduce problematic bugs (see https://github.com/mjmlio/mjml/pull/1058#issuecomment-373749352). 

Also, if what you want to do is give the illusion that the email is plain-text, I don't see why you need MJML? Why not just send the email as plain-text or with a simple HTML layout (`p`, `h1`...)?

thomaschaaf:
I am working on a fix so that the sections don't get put into the same line.

We are using mjml for every email currently. Some emails are automated mails and may look like automated emails. With these emails you are able to also structure some data very well.
However some emails are triggered by our operators. Customers for some reason do not reply to these styled emails but we do use some elements in the emails such as mj-classes (to make the footer always look the same), buttons (for clear call to action items) and columns to structure some elements.

packages/mjml-section/src/index.js
```diff
@@ -100,7 +100,7 @@ export default class MjSection extends BodyComponent {
       },
       div: {
         ...(fullWidth ? {} : background),
-        Margin: marginAlignment,
```
iRyusa: If it gets merged, you should revert this change. Some clients removes some css attributes depending on the case.

thomaschaaf: 👌

thomaschaaf:
@ngarnier I have added a "clearfix" using a 100% table.

```<mjml>
  <mj-body background-color="blue">
    <mj-section align="left" background-color="red">
      <mj-column>
        <mj-divider border-color="#F45E43"></mj-divider>
        <mj-text font-size="20px" color="#F45E43" font-family="helvetica">Hello World</mj-text>
      </mj-column>
    </mj-section>
    <mj-spacer />
    <mj-section align="right" background-color="green">
      <mj-column>
        <mj-divider border-color="#F45E43"></mj-divider>
        <mj-text font-size="20px" color="#F45E43" font-family="helvetica">Hello World</mj-text>
      </mj-column>
    </mj-section>
  </mj-body>
</mjml>
```

https://litmus.com/builder/bc22560

thomaschaaf:
@iRyusa did you look at this? 
Here is an email we send for example: https://www.dropbox.com/s/mecw22dofy91smg/Screenshot%202018-04-04%2009.34.47.png?dl=0

Here we mostly use mjml classes, but as I said we would like to not have to differentiate between two types of email classes if at a later point we'd want to completely redesign the mails.

iRyusa:
Hi,

We're still discussing about this internally, and I'm not sure there's a an huge demand to allow this behavior so that's why it's not merged yet. We're aware that there's no way to edit the outlook conditionnal align too on `mj-section`, we'll try to keep you posted about this

iRyusa:
I think we can close this one for now, i'm not a big fan of adding yet another wrapping table, and the demand is pretty low on this side.

I still think we should add something to target this internal outlook align attributes in some way that it's obvious it won't affect the default behavior, and I really dislike adding some client specific attribute :(

PR #1110, opened 2018-03-30
Fix 1065

kmcb777:


packages/mjml-cli/src/client.js
```diff
@@ -163,7 +176,7 @@ export default async () => {
     // eslint-disable-line array-callback-return
     console.error(`${file ? `File: ${file}\n` : null}${error}`) // eslint-disable-line no-console
 
-    if (config.stack) {
+    if (config.stack || verbose) {
```
iRyusa: Well I think we can remove `config.stack` if we got `verbose` to avoid duplicate options

PR #1129, opened 2018-04-13
handle css-class on Carousel image and thumbnail

kmcb777:


packages/mjml-carousel/src/CarouselImage.js
```diff
@@ -62,14 +62,17 @@ export default class MjCarouselImage extends BodyComponent {
   renderThumbnail() {
     const { carouselId, src, alt, 'tb-width': width, target } = this.attributes
     const imgIndex = this.props.index + 1
+    const cssClass = this.getAttribute('css-class')
+                   ? `${this.getAttribute('css-class')}-thumbnail`
```
iRyusa: You can use multiple css-class on attribute, we already have this logic somewhere to prefix css classes with outlook. Might need to do an helper for that on core ?

packages/mjml-core/src/helpers/suffixCssClasses.js
```diff
@@ -0,0 +1,6 @@
+export default (classes, suffix) => {
+  return classes
```
iRyusa: If classes is `null` then it will output `'null'` string same for `undefined`

PR #1131, opened 2018-04-15
Mjmlconfig

kmcb777:
Need doc update. Works the same as mjml3


packages/mjml/src/index.js
```diff
@@ -46,4 +49,19 @@ registerComponent(CarouselImage)
...
+try {
+  const mjmlConfig = fs.readFileSync(path.join(process.cwd(), '.mjmlconfig'));
+  const custom_comps = JSON.parse(mjmlConfig).packages
+
+  custom_comps.forEach((compPath) => {
+    const requiredComp = require(path.join(process.cwd(), compPath))
+    registerComponent(requiredComp.default || requiredComp)
+  })
+} catch(e) {
+  if (e.code !== 'ENOENT') {
```
iRyusa: We might need a better error handling for each possible cases ?

packages/mjml/src/index.js
```diff
@@ -46,4 +49,19 @@ registerComponent(CarouselImage)
 
 registerDependencies(require('./dependencies'))
 
+
+try {
+  const mjmlConfig = fs.readFileSync(path.join(process.cwd(), '.mjmlconfig'));
+  const custom_comps = JSON.parse(mjmlConfig).packages
```
iRyusa: camelCase 👎

iRyusa:
Anyway, this should be done on `mjml-core` not on the community `mjml` package

packages/mjml/src/index.js
kmcb777: oups

packages/mjml/src/index.js
kmcb777: Actually ENOENT is the case where .mjmlconfig file does not exist
In this case we don't want to consider this as an error
For any other error we log it, but i haven't identified specific error cases

kmcb777:
moved to mjml-core/index
please forgive my ignorance

jskrzypek:
Shouldn't this functionality be run above, inside the `mjml2html()` function? 

The vscode-mjml extension is expecting it to respect passing an alternate directory as the filePath option for the mjml function.

I can make a PR if you want.

iRyusa:
File path is only for relative include. You might need an option for mjmlconfig path ?

jskrzypek:
Yeah I've been ill the past few days so i haven't had a chance to advance on this, but I realized over the w/e that I forgot to allow for node-modules. I'll update my branch/fork when I'm better.

kmcb777:
@jskrzypek i'll merge this one, you can open a PR into the `next` branch when you'll update your branch 👍

PR #1134, opened 2018-04-15
resolve eslint errors

f3ltron:


f3ltron:
je viens de voir les builds fails alors que les eslint-disable etc sont au bonne endroit pour moi

packages/mjml-migrate/src/migrate.js
```diff
@@ -1,4 +1,5 @@
-import fs from 'fs'
+// import fs from 'fs'
```
iRyusa: You can remove it then 👍

f3ltron: i will do it but about travis error i see ? It's not update or something with the code we push ? Because some errors i see eslint should not pass on it

iRyusa: Might be a version conflict on your side, try to clean anything related to eslint in global ?

packages/mjml-cli/src/client.js
```diff
@@ -179,7 +179,7 @@ export default async () => {
   }
 
   switch (outputOpt) {
-    case 'o':
+    case "o": {
```
iRyusa: I think it should be `'o'` for consistency ?

iRyusa:
Can you use the prettify task, I think there's some unwanted spaces in some places too ?

packages/mjml-cli/src/client.js
f3ltron: It could be both. But best pratice tell for '' i will change it

f3ltron:
I use the prettier but it seem's to change text and create eslint error. Like 

```
console.log(  /* eslint-disable-line no-console */
"my text"
);

after prettier 

console.log(  
"my text" /* eslint-disable-line no-console */
);
```

so after it's causing bug so we have to change prettier config or eslint config choose and tell me ^^

iRyusa:
I think you can disable the console log rules for some files. You can do a minimalist eslintrc for mjml-cli & mjml-migrate to allow console log rule ?

PR #1154, opened 2018-04-23
handle image height

kmcb777:


packages/mjml-image/src/index.js
```diff
@@ -93,11 +94,12 @@ export default class MjImage extends BodyComponent {
   }
 
   renderImage() {
+    const height = this.getAttribute('height')
     const img = `
       <img
         ${this.htmlAttributes({
           alt: this.getAttribute('alt'),
-          height: this.getAttribute('height'),
+          height: height && parseInt(height, 10),
```
iRyusa: parseInt of `auto` default value will fail here ?

PR #1175, opened 2018-04-26
mjml-column: Background property on table is invalid, should have been css style or bgcolor instead

### LINKED ISSUE #1174: mjml-column: Background property on table is invalid, should have been css style or bgcolor instead
MJML generates `<table background="#f2f2f2" ...>` which is [not valid](https://developer.mozilla.org/cs/docs/Web/HTML/Element/table).

What happens in `email.seznam.cz` webmail is that they parse this as an unknown property, find a `#f2f2f2` value which is a valid url (unexpectedly and really :) ), assume that it is an url of a background image, and display a warning about email containing external images.

**Reproduction Steps:**

1. Use `<mj-column background-color="#f2f2f2">` in an email template
2. `<table background="#f2f2f2" ...` is generated in the output
3. Email HTML becomes invalid, and some clients understand this construction as providing external image as background, and display warning about external images.

**Expected behavior:**

We use (deprecated) `<table bgcolor="#f2f2f2" ...` instead, or, even better we listen to the advice in [table spec](https://developer.mozilla.org/cs/docs/Web/HTML/Element/table):

> The `<table>` element should be styled using CSS. Use the background-color property in CSS to create an effect that is similar to the bgcolor attribute.

**Observed behavior:**

`email.seznam.cz` webmail displays warning about external images when a `mj-column` with `background-color` is used.

**MJML version:**

4.0.5

**Email clients the bug is seen on:**

email.seznam.cz (webmail)

tkafka:
Partially fixes #1174.

As I don't know the codebase, I did what I could - replacing non-existent `<table>`'s `background` property with deprecated `bgcolor` property in a single place where I know a color will be used - `mjml-column`.

In two other places in `mjml-section` I at least added TODO.

Ideal way of solving this would be to move the background color/image definition into inline css styles in all three places.

### DIFF
--- a/packages/mjml-column/src/index.js
+++ b/packages/mjml-column/src/index.js
@@ -191,7 +191,7 @@ export default class MjColumn extends BodyComponent {
     return `
       <table
         ${this.htmlAttributes({
-          background: this.getAttribute('background-color'),
+          bgcolor: this.getAttribute('background-color'), // TODO: move this attribute into css style as recommended in https://developer.mozilla.org/cs/docs/Web/HTML/Element/table
           border: '0',
           cellpadding: '0',
           cellspacing: '0',
--- a/packages/mjml-section/src/index.js
+++ b/packages/mjml-section/src/index.js
@@ -251,7 +251,7 @@ export default class MjSection extends BodyComponent {
         <table
           ${this.htmlAttributes({
             align: 'center',
-            background: this.isFullWidth()
+            background: this.isFullWidth() // TODO: <table background="..." is not valid and we need to put image bacground there through style, see https://developer.mozilla.org/cs/docs/Web/HTML/Element/table
               ? null
               : this.getAttribute('background-url'),
             border: '0',
@@ -302,7 +302,7 @@ export default class MjSection extends BodyComponent {
         ${this.htmlAttributes({
           align: 'center',
           class: this.getAttribute('css-class'),
-          background: this.getAttribute('background-url'),
+          background: this.getAttribute('background-url'), // TODO: <table background="..." is not valid and we need to put image bacground there through style, see https://developer.mozilla.org/cs/docs/Web/HTML/Element/table
           border: '0',
           cellpadding: '0',
           cellspacing: '0',


packages/mjml-section/src/index.js
```diff
@@ -302,7 +302,7 @@ export default class MjSection extends BodyComponent {
         ${this.htmlAttributes({
           align: 'center',
           class: this.getAttribute('css-class'),
-          background: this.getAttribute('background-url'),
+          background: this.getAttribute('background-url'), // TODO: <table background="..." is not valid and we need to put image bacground there through style, see https://developer.mozilla.org/cs/docs/Web/HTML/Element/table
```
iRyusa: `background` is an old & valid property on `table` element. Some clients like Yahoo doesn't support the `background` property properly and `background` act as a fallback. So please remove both comments.

packages/mjml-column/src/index.js
```diff
@@ -191,7 +191,7 @@ export default class MjColumn extends BodyComponent {
     return `
       <table
         ${this.htmlAttributes({
-          background: this.getAttribute('background-color'),
+          bgcolor: this.getAttribute('background-color'), // TODO: move this attribute into css style as recommended in https://developer.mozilla.org/cs/docs/Web/HTML/Element/table
```
iRyusa: It wasn't here in MJML3 in the first place, can you remove completely this line too please ?

packages/mjml-section/src/index.js
tkafka: Let's tell this in comments then - `// FYI: background is a deprecated property, but some clients like Yahoo doesn't support the background property properly and background property acts as a fallback`

packages/mjml-column/src/index.js
tkafka: Currently it is supported (and it's really useful) - see [mjml-column](https://mjml.io/documentation/#mjml-column), so I'd like to think about how we can preserve this property. Afaik this patch would fix it (as bgcolor is ancient but still supported), right?

iRyusa: https://github.com/tkafka/mjml/blob/472c47c845eb0feebe66a5823a8b6479adbf7f02/packages/mjml-column/src/index.js#L60 Background color is already applied here. A lot of logic has been moved from column to section in MJML4 that's why we still see some unwanted attribute

packages/mjml-section/src/index.js
iRyusa: Well I don't think those comment are mandatory to understand the code, most of the hack we use in MJML are shown either in litmus forum or in this repo https://github.com/hteumeuleu/email-bugs/issues

packages/mjml-column/src/index.js
tkafka: @iRyusa But then I'll lose an option to set background of individual columns (which is a feature I am using now, and a reason I found this problem), right? Or would I then do it with CSS class instead?

iRyusa: > https://github.com/tkafka/mjml/blob/472c47c845eb0feebe66a5823a8b6479adbf7f02/packages/mjml-column/src/index.js#L60 Background color is already applied here.

As said here, no. `background-color` is applied on the wrapping table already so it's just `bgcolor/background` HTML attributes are just useless and color will be applied through the `table` inline style

iRyusa:
Well we fixed the background property so this pr is no longer needed.

Thanks for your contribution

PR #1196, opened 2018-05-14
Fix 82

kmcb777:


iRyusa:
I think we need an attributes for "top" & "bottom" raw ?

PR #1206, opened 2018-05-21
Mjmlconfig

jskrzypek:
@kmcb777 @iRyusa Following up on our discussion from #1131 

I included a couple enhancements that seemed to make sense:
 * you can specify community components installed via npm by their standard requirable package names/paths (i.e. `"mjml-chart"` rather than `"./node_modules/mjml-chart/lib/index.js"`). I handle scenarios where this fails to resolve correctly.
 * if paths in the `packages` array resolve to a non-callable object whose properties are components (like `mjml-accordion` does), I iterate over the keys of the object and register the callable component classes instead

I'm going to create a simple test project and make sure this will work with mjml-charts and an additional custom component and then link the repo here so you guys can verify


packages/mjml-cli/src/client.js
```diff
@@ -83,14 +83,19 @@ export default async () => {
         type: 'object',
         describe: 'Option to pass to mjml-core',
       },
+      cp: {
```
iRyusa: `--config.configPath` will be already defined, I think available options should only be listed in --config option instead

iRyusa:
Hey !

That looks really great for me ! Just I would stick to current `--config` syntax over adding a new option

jskrzypek:
Here's my test repo: https://github.com/jskrzypek/mjml-config-test-project
I'll fix the config option thing now

jskrzypek:
@iRyusa if there's any other changes you'd like me to make on this pr,  just let me know 😄

iRyusa:
HI @jskrzypek 

We just freeze our dev on 4.1 so it will be merged for a 4.1.1 👌

kmcb777:
@jskrzypek nice PR, i'll merge it right now. I'll just change the option name `configPath` into `mjmlConfigPath` and the function `registerCustomComponents` into `handleMjmlConfig` because i also made a PR to make custom components work on the app and that's the names i used ;)

PR #1223, opened 2018-05-28
DOC: Remove duplicated target attribute for mj-social-element

dhuf:


kmcb777:
Nice catch, thanks @dhuf 
I'll merge this after the incoming release, for now merges are frozen

PR #1251, opened 2018-06-21
fix #969 refactor color shorthand hex codes

### LINKED ISSUE #969: MJML should refactor shorthand hex codes
I think MJML should refactor shorthand hex codes to longhand uppercase to prevent issues with some versions of Outlook.

For example:
`#fff` should be changed to `#FFFFFF`

kmcb777:


iRyusa:
Not a big fan to do this on parser side, would prefer to use new type introduced in MJML4.1 and convert on the fly

Issue #200, opened 2016-04-26
Ability to add custom properties


dalefish:
I'd like the ability to add custom properties and values.

Currently I build a lot of editable templates for clients to use in mailchimp, it would be preferable to define the editable areas in the mjml rather than having to retro-fit it to the compiled html.

For instance:
`<mj-text color="#ffffff" font-size="28" align="center" padding-top="40" mc:edit="title">`

dalefish:
Perhaps this format would work?
`<mj-text color="#ffffff" font-size="28" align="center" padding-top="40" prop="mc:edit,title" prop="id,sample_id">`

iRyusa:
I don't know if we can implement this because of how components work. One component doesn't equal one line of html code. So I don't know where to place custom properties...

blackfyre:
This problem seems similar to #324 in their nature.
My recommendation for the the attribute placement would be the outermost html element not encapsulated by `if mso` elements, but it's a tough choice to make...

In case of mailchimp, the [documentation](http://templates.mailchimp.com/getting-started/template-language/) seems to support what I've recommended, but I cannot be certain without actually trying it out.

Example for `mj-text`

```html

<!--[if mso | IE]>
  <table border="0" cellpadding="0" cellspacing="0"><tr><td style="vertical-align:top;width:600px;">
<![endif]-->
  <div id="sample_id" mc:edit="title" aria-labelledby="mj-column-per-100" class="mj-column-per-100" style="vertical-align:top;display:inline-block;font-size:13px;text-align:left;width:100%;">
	<table cellpadding="0" cellspacing="0" width="100%" border="0">
	  <tbody>
		<tr>
		  <td style="word-break:break-word;font-size:0px;padding:10px 25px;" align="left">
			<div style="cursor:auto;color:#F45E43;font-family:helvetica;font-size:20px;line-height:22px;">Hello World</div>
		  </td>
		</tr>
	  </tbody>
	</table>
  </div>
<!--[if mso | IE]>
  </td></tr></table>
<![endif]-->

```

blackfyre:
@iRyusa sorry to bother you, but can you give an ETA on this? Or can I help in any way to speed things up?

iRyusa:
I quote myself, atm we have no idea to implement that :/ 

Each element is unique and has no `root` element. So it's hard to add such generic element.

homerjam:
[...]

PR #1263, opened 2018-07-01
#200 - Ability to add custom properties

bastienrobert:
**WIP, don't merge**

I'm working on the #200 issue, opening this PR to open the discussion about the way to do that.

Personally I thought as @charlex, with that kind of potential solution:
```
<mj-text 
  color="#ffffff" 
  font-size="28" 
  align="center" 
  padding-top="40"
  custom-attrs={
    'mc:edit': 'foobar',
    hello: 'world',
    active: true,
    archived: false,
    'no-value-attr-1': null,
    'no-value-attr-2': undefined,
  }
>
```

So I add a new *object* type, you send an object and inside a string (JSON formatted), so something like that:

```
<mj-button data='{"foo": "bar"}'>
  Hello
</mj-button>
```
(for now, it's only working with button but it's really easy to set on other components)

Data props will be checked by a regex in data object (just like other props types: strings, enum, units...) and then, I parse the JSON into the render method to send it to htmlAttributes method. looks like:

image

If anybody has another idea, perhaps cleaner, I would look 👍

### DIFF
--- a/packages/mjml-button/src/index.js
+++ b/packages/mjml-button/src/index.js
@@ -14,6 +14,7 @@ export default class MjButton extends BodyComponent {
     border: 'string',
     color: 'color',
     'container-background-color': 'color',
+    data: 'object',
     'font-family': 'string',
     'font-size': 'unit(px)',
     'font-style': 'string',
@@ -42,6 +43,7 @@ export default class MjButton extends BodyComponent {
     border: 'none',
     'border-radius': '3px',
     color: '#ffffff',
+    data: {},
     'font-family': 'Ubuntu, Helvetica, Arial, sans-serif',
     'font-size': '13px',
     'font-weight': 'normal',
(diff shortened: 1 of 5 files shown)


iRyusa:
Hi, thanks for this PR. 

As mentioned in the issue there's still no way to set where the custom attributes would be applied :/ 
Personally I don't really find the JSON notation really friendly too here 🤔

bastienrobert:
Yes I know, the major problem is about the syntax. As we discussed in the #200 [here](https://github.com/mjmlio/mjml/issues/200#issuecomment-402053887), if you want to add custom datasets with weird syntax as the mailchimp one, I think it's going to be a problem with react inline props.

iRyusa:
Well for now I'm closing this pr, there's no real solution found yet to the given implem of custom attributes that would satisfy everyone. let's keep the discussion open in the issue.

Issue #1135, opened 2018-04-16
background-attribute of tables is receiving a color (expecting url)


Michael-wm:
The css background-color is assigned to the background-attribute of tables and tds (which is expecting an image-url) by default. Bgcolor might be the correct attribute for the background-color.

Link to the code-culprit:
https://github.com/mjmlio/mjml/blob/354519adaa8cd9cd488c7dc95f67f18c283ed412/packages/mjml-column/src/index.js#L213

https://github.com/mjmlio/mjml/blob/354519adaa8cd9cd488c7dc95f67f18c283ed412/packages/mjml-column/src/index.js#L236

PR #1288, opened 2018-07-26
fix: background-attribute was still causing problems

Michael-wm:
This is regarding the fix for issue #1135 which was partly fixed by this commit: https://github.com/mjmlio/mjml/commit/62d56e1ba28bf1a669ce1f96d4767fcfd483c50a

There is still one occurence of the incorrect background-attribute in the renderGutter method that is still causing problems. When I remove it, everything works fine.


iRyusa:
👍

PR #1293, opened 2018-07-30
Improve documentation around validation levels when using CLI

jordinl83:
The documentation for setting validation level when using the CLI mentioned a `-l` option to set the validation level. This option seems to be no longer supported, but setting `--config.validationLevel=$level` works. I added documentation for the `--config.validationLevel` option in both the [Usage](https://github.com/mjmlio/mjml/blob/master/doc/install.md#command-line-interface) section and the [Validations](https://github.com/mjmlio/mjml/blob/master/packages/mjml-validator/README.md#validating-mjml) section.


iRyusa:
👍 
Thanks about this

PR #1295, opened 2018-07-30
Fix 1281

kmcb777:


packages/mjml-hero/src/index.js
```diff
@@ -126,6 +126,7 @@ export default class MjHero extends BodyComponent {
         height: this.getAttribute('background-height'),
         'mso-position-horizontal': 'center',
         position: 'absolute',
+        top: 0,
```
iRyusa: top: '0' and the `isNil` is no longer needed ?

kmcb777: but i don't see why `0` should be ignored and `1` wouldn't be ignored. There are other styles in mj-hero that have the value `0` just above and i don't think they were supposed to be ignored. I.e. somebody making a custom component would have a hard time understanding why this is ignored

iRyusa: 👍

PR #1315, opened 2018-08-15
Update `ignoreInclude` flag to be `ignoreIncludes`

zackharley:
When creating a new component, the MJMLParser does not ignore includes due to the flag being named incorrectly.


kmcb777:
hi @zackharley thanks, nice catch

zackharley:
@kmcb777 No problem. Happy to help

PR #1328, opened 2018-08-23
fix #359 full clickable button

### LINKED ISSUE #359: Entire button as link

kmcb777:


hieunguyen135:
@kmcb777 @iRyusa tested, it's ok https://litmus.com/checklist/emails/public/4843ca9

PR #1337, opened 2018-08-31
Remove illegal text-mode attribute in documentation

gileri:
`text-mode` [is unused when rendering `<mj-social-element>`](https://github.com/mjmlio/mjml/blob/master/packages/mjml-social/src/SocialElement.js#L194-L207), and it is deemed illegal during [validation](https://github.com/mjmlio/mjml/blob/master/packages/mjml-social/src/SocialElement.js#L46-L68).


kmcb777:
@gileri that's true, thanks for noticing this

gileri:
Thank you for the quick merge !

PR #1349, opened 2018-09-13
revert fix that prevents images from shrinking on mobile

kmcb777:


kukac7:
```
mjml-core: 4.2.0-beta.2
mjml-cli: 4.2.0-beta.2
```

the image's responsiveness does not work on mobile.

iRyusa:
Yes, we're aware of that. That's why we revert this change in 4.2, new beta should come this week.

PR #1366, opened 2018-10-05
Fix mj-spacer default height

timmw:
Docs say `mj-spacer` should have a default height of 20px but no default is set. 

https://mjml.io/documentation/#mjml-spacer


timmw:
Oops, this was already fixed in #1313

Issue #1367, opened 2018-10-05
`background` property on the table contains invalid value


mastertheblaster:
The following mjml:

```html
<mjml>
  <mj-body>
    <mj-section>
      <mj-column padding="10px" background-color="red">
        <mj-divider border-color="#F45E43"></mj-divider>
      </mj-column>
    </mj-section>
  </mj-body>
</mjml>
```

creates HTML with a `table`, having property `background` which is equal to the specified color.
In this case `<table background="red"/>`. `background` property is used to specify the image, not the color. Color should be specified using the `bgcolor` property.

Is this a workaround for some issue or just by accident?

iRyusa:
Hi,

Yeah this should be removed and the `container-background-color` one too.

iRyusa:
It seems that it has been fixed in recent 4.2 release, so i'm closing

PR #1368, opened 2018-10-05
fix(mj-column): change background to bgcolor

mastertheblaster:
Should solve issue #1367


iRyusa:
Hi, thanks for the PR,

We just have to remove this one, it's not totally useless, can you remove this ?

We'll merge it for 4.3

iRyusa:
👋 any news on this ?

iRyusa:
👋 
We did remove this `background` attribute because it has no real purpose 
https://github.com/mjmlio/mjml/blob/master/packages/mjml-column/src/index.js#L192-L194

PR #1378, opened 2018-10-10
 fix: removeContainerTag function

7170:
Here is a fix for the removeContainerTag() function which was removing any first tag instead of only a mj-container one.

Could you please review it and merge it if it meets your requirements?


iRyusa:
👍 Thanks for this fix on the migrate

Issue #1385, opened 2018-10-15
[Documentation] Order attributes in arrays


totocap:
Attributes are not always in alphabetical order in README.md. The search of an attribute in the arrays takes more time than necessary.

MJML Version : 4.2.0

![image](https://user-images.githubusercontent.com/10602197/46936903-31752a00-d060-11e8-8bb5-4553347bf667.png)

PR #1386, opened 2018-10-15
Doc #1385 : order attributes in array

totocap:
- Reorder all attributes in each README.md of each package so they can
be in alphabetical order.


iRyusa:
👍 Thanks for this PR, we'll try to rebuild the live doc asap

PR #1454, opened 2018-12-14
fix link to "try it live"

murtho:


kmcb777:
Hi @murtho thanks for fixing this, full link is indeed needed when viewing on github 👍

PR #1462, opened 2018-12-19
Completely flush the rendered html to stdout before exiting in mjml-cli

### LINKED ISSUE #1461: The mjml-cli command exits before flushing all output to stdout
**Describe the bug**

The `outputToConsole` helper in the `mjml-core` package writes to stdout using an asynchronous `console.log()` call and then immediately resolves regardless of whether the data is completely flushed or not. When using the `mjml` command with the `-s` option and piping the result into another command, this can result in the process exiting before all of the data is flushed through the pipe. On Linux, this results in the truncation of the piped output after the first 65536 bytes (corresponding to the [maximum pipe capacity](https://linux.die.net/man/7/pipe)).

**To Reproduce**

1. Create a file with MJML code that is expected to render to more than 65536 bytes. This can be anything as long as the output is large. Here's a command that will generate one:

        echo "<mjml><mj-body><mj-text>$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 100000 | head -n 1)</mj-text></mj-body></mjml>" > /tmp/template.mjml

2. Render it to HTML and count the number of bytes by running:

        cat /tmp/template.mjml | ./node_modules/mjml/bin/mjml -i -s | wc -c

3. Observe that the number of bytes piped into the `wc` command is 65536.

**Expected behavior**

The full rendered template should be flushed to stdout before the process exits and the pipe closes. The number of bytes piped into `wc` should therefore be larger than 65536.

 - OS: Linux
 - MJML Version: 4.2.1
 - MJML tool used: mjml-cli

 - N/A

**Affected email clients (for rendering issues):**
 - N/A

**Screenshots**
 - N/A

**Additional context**
 - N/A

sangaline:
This fixes #1461 by switching from the non-blocking `console.log()` call in `outputToConsole` to a `process.stdout.write()` call. The `outputToConsole` promise would previously resolve immediately, but it will now resolve only after the output has been completely flushed to stdout.

Closes #1461


kmcb777:
Hi @sangaline thanks for reporting this, and providing the solution !
We will merge it in the next version

Issue #1479, opened 2019-01-08
Alt text hidden in most email clients


Zelltron:
**Describe the bug**
Mj-image is inheriting a parent font-size of 0 from mj-section by default. This reduces the font-size of alt text in unloaded/broken images. Alt text is hidden in most email clients. 

**To Reproduce**
1. Create a file with this MJML code: 
```
<mjml>
  <mj-head>
    <mj-attributes>
      <mj-text align="center" color="#555" />
      <mj-section background-color="#fff" />
    </mj-attributes>
  </mj-head>
  <mj-body background-color="#eee">
    <mj-section>
    <mj-column>
        <mj-image width="200px" src="http://placehold" alt="testing"></mj-image>
      </mj-column>
    </mj-section>
  </mj-body>
</mjml>
```
2. Render it to HTML by doing either...
- ` $ mjml --watch input.mjml --output index.html`
- Generate the HTML through a native app 
- The MJML online editor

3. Send the HTML to an email address with gmail, apple mail, or yahoo mail

**Expected behavior**
The image alternate text displays over blocked image.

**MJML environment (please complete the following information):**
 - OS: Sierra 10.12.6 
 - MJML Version 4.3.0
 - MJML Node Module

**Email sending environment(for rendering issues)**:
 - Litmus

**Affected email clients (for rendering issues):**
 - Emails: Client Gmail, Yahoo, Apple Mail
 - OS: Windows/Apple
 - Browsers: Chrome & Firefox

iRyusa:
Well it can be styled through mj-style inline & some css selector, but you're right we should expose mj attributes like font-size/color on mj-image

Zelltron:
This worked, but I had to use `!important` on font-size to view the alt text.

iRyusa:
This has been merged for a while now :D

PR #1483, opened 2019-01-10
Fix Alt Text Hidden Bug #1479/Allow for font-size && Allow max-height to Ch…

Zelltron:
Fix Alt Text Hidden Bug [#1479](https://github.com/mjmlio/mjml/issues/1479)/Allow for font-size && Allow max-height to Change Placeholder Size

I've fixed the bug hiding alt text by default for images. This also allows users to change the font-size if they want to. I've also added the ability for users to change the default size of blocked images in their emails. This is in case a user wants the email to look more accurate to the fully loaded content.


iRyusa:
For now @chris13524 you can use mj-style + inline as a workaround, we'll try to release this with 4.4.0 as soon as we can

chris13524:
I'm not able to get mj-style + inline to work with images (I'm new to MJML btw).

```
<mjml>
  <mj-head>
    <mj-style inline="inline">
      .logo {
        font-size: 22px;
      }
    </mj-style>
  </mj-head>
  <mj-body>
    <mj-section>
      <mj-column>
        <mj-image css-class="logo" alt="Alt text."
                  src="https://www.google.com/images/branding/googlelogo/1x/googlelogo_color_272x92dp.png.nope"></mj-image>
      </mj-column>
    </mj-section>
  </mj-body>
</mjml>
```

iRyusa:
`.logo img` the css-class is applied to the first element generated, not the last one

PR #1485, opened 2019-01-11
Add css-class to body components documentation

gileri:
The `css-table` attribute is accepted by all body components.


iRyusa:
👍 thanks for this

Issue #1514, opened 2019-02-15
<mjml-spacer> - border


OReshetnikov:
The documentation for the <mjml-spacer> enumerate attributes which we can use (please see -> https://mjml.io/documentation/#mjml-spacer)
But something wrong with border attribute. Looks like attribute is broken or I'm doing something wrong:)

I try to use it like: 
```
<mjml>
  <mj-body>
    <mj-section>
      <mj-column>
        <mj-text>A first line of text</mj-text>
        <mj-spacer height="30px" container-background-color='#f3efef' border='10px solid red' padding='10px 20px'></mj-spacer>
        <mj-text>A second line of text</mj-text>
      </mj-column>
    </mj-section>
  </mj-body>
</mjml>
```
And.... nothing happens.

How to resolve this issue?

iRyusa:
Hi @OReshetnikov 
It looks like an error in the doc, spacer doesn't support any border attribute. If you can do a PR about this one it would be great ! 👍 

You can use mj-style to apply a border on the td, the support should be fine

PR #1518, opened 2019-02-18
update mjml-spacer documentation (remove redundant border attributes)

OReshetnikov:
According to the issue https://github.com/mjmlio/mjml/issues/1514, redundant border attributes were deleted from documentation


iRyusa:
👍 Thanks

PR #1531, opened 2019-02-28
Allow 2 values for inner-padding on mj-button

aew:
Providing 2 padding numbers (for top/bottom and left/right) is valid for HTML `td` tags, and also the [default value](https://github.com/mjmlio/mjml/blob/master/packages/mjml-button/src/index.js#L49).  I believe this change will allow the validator to accept 2 values without a warning like `(mj-button) — Attribute inner-padding has invalid value: 15px 90px for type Unit, only accepts (px, %) units and 1 value(s)`


aew:
Noticed that this is already fixed in the `next` branch

raphaelfabeni:
Is this working? I'm using the _4.4.0-beta.1_ and I got the same error.

iRyusa:
This is not released yet. But it does support it anyway it’s just à validation issue.

PR #1554, opened 2019-03-26
fix #1527 doc to ignore content during minification

### LINKED ISSUE #1527: HTML minification parser error
**Describe the bug**
HTML minification breaks if there is  a `<` in a `<mj-raw>` section.

**To Reproduce**
```
<mjml>
  <mj-body>
    <mj-section>
      <mj-column>
        <mj-raw>
          {% if foo < 5 %}
          {% endif %}
        </mj-raw>
        <mj-text>Hello World</mj-text>
      </mj-column>
    </mj-section>
  </mj-body>
</mjml>
```
2.  compile using the API:
```
const output = mjml2html(file.contents.toString(), {
  minify: true,
  filePath: file.path,
})
```
3. See error: 
```
Error: Parse Error: < 5 %}
```

**Expected behavior**
Expected minified HTML to succeed

 - OS: MacOS 10.14.3
 - MJML API Version 4.3.1

kmcb777:


iRyusa:
Maybe let's add an example bloc on the doc ?

PR #1570, opened 2019-04-09
Change 'mjml-' to 'mj-' (1) everywhere it refers to a component or ta…

GarryFlemings:
…g (relevant to MJML users) and (2) nowhere it refers to a package (relevant to MJML programmers).

2019-04-08: Trying to give you access to my inputs.


GarryFlemings:
2019-04-12: Adding inputs to resolve Issue #1574.

packages/mjml-cli/README.md
```diff
@@ -77,28 +77,12 @@ $> mjml input.mjml --config.minifyOptions='{"minifyCSS": true, "removeEmptyAttri
 
 The defaults are "collapseWhitespace": true, "minifyCSS": false, "removeEmptyAttributes": true
 
-### Preserve specific tags when using inline mj-style
```
iRyusa: Well this block should stay

packages/mjml-cli/README.md
```diff
@@ -77,28 +77,12 @@ $> mjml input.mjml --config.minifyOptions='{"minifyCSS": true, "removeEmptyAttri
...
-
-When using `<mj-style inline="inline">` the css will be inlined using the juice library. As a side effect, juice will convert all tags' attributes into lower case. If you need to preserve some cases (i.e. for a templating lib) you can specify the tags to preserve. With the example above, all tags of the form `<# myVar="" >` or `</# myVar="" >` will be left untouched. By default juice already ignores `<% EJS %>` and `{{ HBS }}` tags.
-
 ### Log error stack
 
 ```bash
 $> mjml input.mjml --config.stack true
 ```
 
-### Override base path for mj-include relative paths
```
iRyusa: This block should stay too

packages/mjml-raw/README.md
```diff
@@ -19,19 +19,3 @@ If placed inside `<mj-head>`, its content will be added at the end of the `<head
   </a>
 </p>
 
-
```
iRyusa: this should stay

packages/mjml-group/README.md
```diff
@@ -61,5 +61,5 @@ attribute           | unit        | description                    | default att
 width               | percent/px  | group width                    | (100 / number of non-raw elements in section)%
 vertical-align      | string      | middle/top/bottom              | top
 background-color    | string      | background color for a group   | n/a
-direction           | ltr / rtl   | set the display order of direct children | ltr
```
kmcb777: this should stay too

packages/mjml-section/README.md
```diff
@@ -44,12 +44,12 @@ border-radius       | px          | border radius                  | n/a
 border-right        | string      | css border format              | n/a
 border-top          | string      | css border format              | n/a
 css-class           | string      | class name, added to the root HTML element created | n/a
-direction           | ltr / rtl   | set the display order of direct children | ltr
```
kmcb777: this should stay unchanged

kmcb777:
@GarryFlemings thank you for this PR 👍 however it seems like recent changes to the doc made by other commits have been reverted, maybe during conflict resolution
It would be great if you can fix these !

packages/mjml-cli/README.md
GarryFlemings: Sorry for taking so long to respond. I aspire to answer faster.

I have no reason to disagree with any of these. At the same time, I recommend careful review of my output.
1) I don't recall having made these changes.
- Maybe the second blank line at the end of mjml-raw/README.md. I didn't know that was important. I had some changes not directly related to `mjml-` to `mj-` in a draft of my PR, but intended to delete all those. I should have deleted this one.
2) I've compared the current `next` to `9d9e2454` (the commit immediately before my release). Where (like the above `### Preserve specific tags when using inline mj-style`) is absent from 'next' and in '9d', I'm doing copy-and-paste to put the '9d' version in my commit. If there has been a later change to any of the copy-and-paste text, I won't catch it. Maybe there's a better way to recover from this condition. (Don't know about that.)

Action: Add the `9d` version of the `### Preserve specific tags when using inline mj-style` block to `next`.

Note: Either I missed a change in `mjml-button/README.md` or was lost.

Action: Changed `mjml-button` to `mj-button` on line 1, consistent with the intent of the original PR.

packages/mjml-cli/README.md
GarryFlemings: Action: Copied `9d` version of `### Override base path for mj-include relative paths` block to `next`.

packages/mjml-raw/README.md
GarryFlemings: No action. This second blank line is absent in `9d`. It is present in `next`. Perhaps someone else corrected this?

packages/mjml-group/README.md
GarryFlemings: Added the `9d` version of `direction` to `next`.

packages/mjml-section/README.md
GarryFlemings: Action: Substituted the `9d` version of `direction` for the `next` version.

* An aside: Both forms are technically correct ("ltr" and "rtl" are both strings). I consider the above form preferable because only certain strings are meaningful. There are other places in the documentation with this condition. I'll do a PR for those at some point and see how the other ideas appeal.

GarryFlemings:
I pushed new commit `722c1d2f` to `GarryFlemings:next` on GitHub containing the above changes.

packages/mjml-raw/README.md
iRyusa: I meant the whole block after that not the blank line :)

packages/mjml-section/README.md
```diff
@@ -52,3 +52,4 @@ padding-left        | px          | section left offset            | n/a
 padding-right       | px          | section right offset           | n/a
 padding-top         | px          | section top offset             | n/a
 text-align          | string      | css text-align                 | center
+vertical-align      | string      | css vertical-align             | top
```
kmcb777: this had been removed intentionally so it shouldn't be re-added

packages/mjml-group/README.md
kmcb777: ok for the `direction` doc changes 👍 it's fine like this indeed

packages/mjml-section/README.md
GarryFlemings: commit 3b9a67e1

packages/mjml-raw/README.md
iRyusa: Bump on this block, this is an important thing to keep for 4.4

GarryFlemings:
PR #1764 contains all changes intended for this PR.

GarryFlemings:
PR#1570 includes these changes.

PR #1573, opened 2019-04-11
move cross-env to dev dependency

### LINKED ISSUE #1572: Why is cross-env a runtime dependency?
Particularly when it comes to the `mjml-core` package, `cross-env` appears to be used at build-time only (via `npm run build` or `npm run clean`) - there are no references to `cross-env` in production code distributed in the published package.

As an application developer using mjml, I'm looking for ways to keep our dependency footprint (and thus, our security audit reports) as small as possible.

Are you open to moving `cross-env` to be a dev dependency instead?

nexdrew:
Fixes #1572.

This PR is an addendum to #1091 and makes the following changes:

1. Removes `cross-env` as a dependency from the parent (mjml-master) package.json, since it is unnecessary there (unable to share the parent dependency with all child mjml packages)
2. Moves `cross-env` from a runtime dependency to a dev dependency for all child packages, since it is only needed at build-time
3. Bumps the version of `cross-env` used from 5.1.4 to 5.2.0

Note that when I first got the project set up on my local machine, there was a change made to **mjml-cli/yarn.lock** just by running `yarn`. I captured this change in a separate branch and commit, which are included in this PR, but you can view it independently here: https://github.com/mjmlio/mjml/compare/master...nexdrew:mjml-cli-yarn-lock

Also note that we appear to be testing against older versions of Node in Travis; currently .travis.yml is configured to use Node 6, 8, and 9, which doesn't even include the current active Node LTS (10.15.3). At the end of this month, Node 6 will be EOL and Node 12 will be released. We might want to consider modifying Travis config to drop Node 6 and replace Node 9 with `lts/*` to always test on latest Node LTS.

Also note that this project appears to be using an old version of Lerna, and we might want to consider migrating from Lerna 2 to Lerna 3.

Let me know if you have any questions or concerns.


iRyusa:
👍 for bumping node version to lasts LTS

It's really weird that all those package have now way more deps (in lock file 🤔) I wonder if we should move mjml-core as a peer-dep instead of deps on those components packages ?

kmcb777:
an other PR was opened with these commits

PR #1589, opened 2019-04-25
Added missing documentation for mjml-table

Busteren:
css-class was missing in table documentation
also moved css-class in mjml-text to be more consistent with other documentation


kmcb777:
@Busteren Thanks for this PR 👍

kmcb777:
@Busteren aw, actually we always merge in the `next` branch first, then into master for new releases.
Here changing the base branch to `next` causes unwanted changes, would you mind making these changes again, but starting from `next` ? It should be easier than fixing this one

Busteren:
@kmcb777 had a look in the next branch and it contains the changes🙌

Issue #579, opened 2017-03-04
How to : Icon + Text


maximeparisse:
Hi.

I've some hard time trying to do something simple in non-responsive emailing.

![Example](https://i.imgsafe.org/a8d095e115.png)

How would achieve to do this in a responsive way ?
The requierements would be:
- The icon size should be fixed, let's say 40x40
- The icon shouldn't go above/below the text in mobile's versions
- I would like to use raw html as less as possible

In a non-responsive way it would be easy to just make 3 columns with fixed widths, but using mjml i don't manage to have something that can work in both mobile/desktop versions. Can you help me about that ?

maximeparisse:
After some deeper searches, i suppose the best way for me to accomplish that would be to use the mj-raw component. Something like this : https://mjml.io/try-it-live/rJWyXF_5e

Am i right ?
Of course it should be more emailing-way, but that's the main idea.

ngarnier:
Hey @maximeparisse, there's a simple way to achieve a responsive layout like that, using 2 `mj-column` wrapped in a `mj-group`, which purpose is to prevent columns from stacking: https://mjml.io/documentation/#mjml-group (note the comment about iOS in the doc).

If you want exact widths rather than relative, using HTML might be a solution, note that you can: 
- use `mj-table` to include a table: https://mjml.io/documentation/#mjml-table
- use `mj-text` that supports any HTML

I'm closing as I think this should answer your question, but feel free to answer if it doesn't.

PR #1663, opened 2019-08-06
Allow social elements without links

krystof-k:
Opening this as a workaround for #579. Social elements work as a perfect media object (fixed width element on the left + text filling the remaining width).

It will be possible to rewrite this into a completely new component but this is out of my abilities at the moment. I know it isn't an ideal solution so feel free to close it if this is something that should not be merged – however I think it doesn't break anything and actually works just fine.

This is what I want to achieve:

image

If there is any cleaner workaround just tell me :)

### DIFF
--- a/packages/mjml-social/src/SocialElement.js
+++ b/packages/mjml-social/src/SocialElement.js
@@ -229,11 +229,13 @@ export default class MjSocialElement extends BodyComponent {
           >
             <tr>
               <td ${this.htmlAttributes({ style: 'icon' })}>
-                <a ${this.htmlAttributes({
-                  href,
-                  rel: this.getAttribute('rel'),
-                  target: this.getAttribute('target'),
-                })}>
+                ${this.getAttribute('href') != '' ?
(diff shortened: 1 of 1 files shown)


kmcb777:
Hi @krystof-k , thanks for this PR

@iRyusa what do you think ? It's not a "social"-related feature, but people have a recurrent need of putting icons and text side by side, it would be much easier for them to use this than mj-table.
Do you think that completely removing the `a` tag can break the display ?

However i think we need a different check than `this.getAttribute('href') !== ''`, we should check if it's been defined at all

iRyusa:
We already do the same with `mj-button` ... if there's no href we use `p` instead of `a`.  

But I'm with you too mj-table is a bit "better" for that, but it's like 5 times more verbose than using  mj-social.

iRyusa:
Can you rebase this on `next` branch please @krystof-k  ?

krystof-k:
Sure, done.

PR #1673, opened 2019-08-08
Update to latest Yargs to resolve vulnerable transient dependencies

hermanbanken:
GitHub reports a CVE in the following dependency chain:

mjml@4.4.1 > mjml-cli@4.4.1 > yargs@^8 > os-locale@^2 > mem@^1

All of these are patch pinned, and mem@1 is not patched, however,
both yargs@8 and mem@1 are WAY behind: yargs@13 and mem@4 are available.
I suggest mjml-cli updates to yargs@^13 to resolve this vulnerability
for everyone depending on mjml.

### DIFF
--- a/packages/mjml-cli/package.json
+++ b/packages/mjml-cli/package.json
@@ -28,7 +28,7 @@
     "mjml-migrate": "4.4.1",
     "mjml-parser-xml": "4.4.1",
     "mjml-validator": "4.4.1",
-    "yargs": "^8.0.2"
+    "yargs": "^13.3.0"
   },
   "devDependencies": {
     "chai": "^4.1.1",


iRyusa:
Hey thanks for this PR !

Did you check that nothing break here, it seems like a big jump ?

hermanbanken:
Short answer: no.
Normally one would have tests for that, but I don't see any in this package. I'm not an active user of this project, I just happen to depend on mjml, which itself has a dependency on its cli.

iRyusa:
Yeah the test coverage on MJML is .... really low...  🙃 

There's a list of command on the main readme : https://github.com/mjmlio/mjml#command-line-interface

We only use yargs to handle most of input options so it shouldn't be that hard to check if everything is ok

Buuhuu:
As part of #1700 I tested a few of the comments and it looks like mjml's use of yargs is not effected by the breaking changes.

PR #1680, opened 2019-08-12
Support custom attributes for mj-image link

tianhsky:
Story: some third party email delivery service (such as Campaign Monitor) can modify links to support their own link tracking, certain behavior can be controlled by supply custom attributes to the anchor tag.
Solution: this commit introduces new attribute to mj-image, here is how to use it <mj-image custom-link-attrs="key1:val1" />, it will add key="val1" to the a tag.

### DIFF
--- a/packages/mjml-image/src/index.js
+++ b/packages/mjml-image/src/index.js
@@ -111,7 +111,7 @@ export default class MjImage extends BodyComponent {
     `
 
     if (this.getAttribute('href')) {
-      let htmlAttrs = {   
+      const htmlAttrs = {   
         href: this.getAttribute('href'),
         target: this.getAttribute('target'),
         rel: this.getAttribute('rel'),
--- a/packages/mjml-image/src/index.js
+++ b/packages/mjml-image/src/index.js
@@ -34,7 +34,8 @@ export default class MjImage extends BodyComponent {
     height: 'unit(px,auto)',
     'max-height': 'unit(px,%)',
     'font-size': 'unit(px)',
-    'custom-link-attrs': 'string',
+    'custom-attr-a': 'string',
+    'custom-attr-img': 'string',
   }
 
   static defaultAttributes = {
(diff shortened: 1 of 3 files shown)


iRyusa:
Well, the issue with custom attributes (and with your implementation too) is : you can't target any other tag. 

This has been discussed a bit here #200  (cf https://github.com/mjmlio/mjml/issues/200#issuecomment-471979086 ).

packages/mjml-image/src/index.js
```diff
@@ -34,6 +34,8 @@ export default class MjImage extends BodyComponent {
     height: 'unit(px,auto)',
     'max-height': 'unit(px,%)',
     'font-size': 'unit(px)',
+    'custom-attr-a': 'string',
```
iRyusa: And then we add `custom-attr-tr`, `custom-attr-td`, custom-attr-XX, this won't really scale really much :/

tianhsky: custom-attr-[tag] will allow user to specify which tag to attach the attr to, which gives the user some control over it. Adding attr to every tag is probably not necessary.

iRyusa: There's a lot of HTML produced by a single tag an the current solution would require to do for every single html element output by mjml, even (and for like... section it would be hard to do so because you have multiple table/tr/td/....).

tianhsky: The point is, in most cases, developer would add custom attrs to a/img/p tags, table/tr/td are for getting the view display right for different mail clients, so I think custom attr can be supported for a selected list of tags, at least the most common ones.
What would you say is the idea way to assign custom attrs?

iRyusa: If we just apply them for a small and restricted list of tags, it has no real benefit.

I have a small idea on how to implement this correctly, i'll post it soon in the custom attribute issue.

tianhsky: Another way to do this  systematically is to add a wrapper to htmlAttributes to so that it can access more context, then it will know what kind of tag it is, and is there a custom-attr-[tag] for the subject tag, if so, it can attach the attr to it. This will require a bigger refactor. Let me know what you think about this approach.

iRyusa: I think that can be done too, but then it still require to add them 1 by 1 let's continue that conversation in #200 👍

PR #1700, opened 2019-09-09
Updated yargs to 13.0.0

Buuhuu:
There has been a a critical security issue reported for one of the transitive
dependencies in yargs@8.0.0. Namely it is the following dependency tree:

yargs@8.0.x => os-locale@3.0.0 => mem < 4.0.0

The reference to mem@4.0.0 as been updated in >= os-locale@3.0.1 which is
in place for yargs since 13.0.0

See https://github.com/yargs/yargs/pull/1270


Buuhuu:
Duplicates #1673

PR #1703, opened 2019-09-10
fix #1689 add warning when included file not found

### LINKED ISSUE #1689: Bad path in mj-include does not make a validation error
**Describe the bug**
If you use mj-include, when you call eg.
```
    const mostlyRenderedHTML = mjml2html(mjmlTemplate);
```

you might get errors with not being able to find the includes as comments in the html, but they should be validation errors IMO as they represent a serious ommission with the rendered html.

**To Reproduce**
Have mj-includes in a subdirectory.

**Expected behavior**
`mostlyRenderedHTML.errors` should not be `[]`, or strict validation should throw an error.

 - OS: Linux
 - MJML Version 4.2
 - MJML tool used node package

kmcb777:


packages/mjml-parser-xml/src/index.js
```diff
@@ -74,6 +74,9 @@ export default function MJMLParser(xml, options = {}, includedIn = []) {
         tagName: 'mj-raw',
         content: `<!-- mj-include fails to read file : ${file} at ${partialPath} -->`,
         children: [],
```
iRyusa: why not `error: { type: 'include',  params: {file, partialPath}}` here

packages/mjml-validator/src/rules/errorAttr.js
```diff
@@ -0,0 +1,14 @@
+import { get } from 'lodash'
+import ruleError from './ruleError'
+
+export default function erroAttr(element, { components }) {
```
iRyusa: And here check for `element.errors` ? (function name has a typo too)

packages/mjml-parser-xml/src/index.js
kmcb777: yes better indeed to keep this separated from attributes

PR #1704, opened 2019-09-10
fix #1639 accept negative values

### LINKED ISSUE #1639: Negative value of "letter-spacing"
**Describe the bug**
When you use negative value of "letter-spacing" you will see this message: "Attribute letter-spacing has invalid value: -1px for type Unit, only accepts (px, %) units and 1 value(s)".

**To Reproduce**
1. Create a MJML text component with `letter-spacing="-1px"`
2. Render it to HTML.

**Expected behavior**
No error message.

 - OS: MacOS 10.13.6
 - MJML Version: 4.3.0
 - MJML tool used: Visual Studio Code

kmcb777:


packages/mjml-core/src/types/unit.js
```diff
@@ -21,7 +21,7 @@ export default params => {
 
       this.matchers = [
         new RegExp(
-          `^(((\\d|,|\\.){1,}(${filteredUnits.map(escapeRegExp).join('|')})|0${allowAuto})( )?){${args.join(
+          `^(((-|\\d|,|\\.){1,}(${filteredUnits.map(escapeRegExp).join('|')})|0${allowAuto})( )?){${args.join(
```
iRyusa: Add an option for negative unit support ?

kmcb777: an option at which level ? You would like us to only allow negative values for specific attributes ?

iRyusa: At the type level, it should allow only positive or positive + negative value.

mj-column width="-600px" is valid with this change, it shouldn't

PR #1705, opened 2019-09-10
fix #1628 filePath accepts directories

### LINKED ISSUE #1628: CLI Issue: --config.filePath does not work as expected
## Describe the bug

Using the CLI, using `--config.filePath` flag that dictates where `mj-includes` are referenced from, meaning you can use partials in a separate directory.

However, the flag currently expects a **single file**, not a **directory**

### Expected behavior: 

`mjml --c.filePath includes -w src/emails/transactional/test-2.mjml -o`

Should generate an email using the `/includes/` directory. **This currently fails.**

### What does work:

`mjml --c.filePath includes/header.mjml -w src/emails/transactional/test-2.mjml -o`

While this works, it's **not clear** that this works for **multiple partials**. (however, it's also error prone when working with many emails)

### To Reproduce

Test Repo: https://github.com/araphiel/mjml-filePath-issue

You can clone this repo and run the following commands to see odd but working implementations.

 - OS: MacOS
 - MJML Version: 4.4.0-beta.2

Related:
https://github.com/mjmlio/mjml/issues/1546
https://github.com/mjmlio/mjml/issues/842

kmcb777:
Since we made 'filePath' a public option to change includes base path, we need to handle it correctly. It was still used internally, i.e. for validation messages, and this was now incorrect.
Since the cli handle watching nested includes, i think it's valuable to allow to provide a directory instead of a file, so this is now handled.


packages/mjml-parser-xml/src/index.js
```diff
@@ -43,7 +44,11 @@ export default function MJMLParser(xml, options = {}, includedIn = []) {
     map(component => component.getTagName()),
   )({ ...components })
 
-  const cwd = filePath ? path.dirname(filePath) : process.cwd()
+  let cwd = process.cwd()
+  
+  if (filePath) {
+    cwd = path.extname(filePath) === '' ? filePath : path.dirname(filePath)
```
iRyusa: Shouldn't we just check if it's a directory here instead ?

Issue #1671, opened 2019-08-07
Hero background image is under wrapper background in Outlook


krystof-k:
When using _mj-hero_ inside _mj-wrapper_, which has a background color, the hero background (image) is hidden below the wrapper background color in Outlook because of `z-index: -3;`:

https://github.com/mjmlio/mjml/blob/11c88d13e827301fed34c446283f7aff6d867a83/packages/mjml-hero/src/index.js#L129

**To Reproduce**
1. Create a file with this MJML code:

   ```mjml
   <mjml>
     <mj-body>
       <mj-wrapper background-color="pink">
         <mj-hero
           mode="fixed-height"
           height="300px"
           background-width="600px"
           background-height="300px"
           background-url="https://placekitten.com/600/300"
           background-color="red"
         >
           <mj-text
             color="#ffffff"
             align="center"
             font-size="45px"
             line-height="45px"
             font-weight="900">
             MEOW
           </mj-text>
         </mj-hero>
       </mj-wrapper>
     </mj-body>
   </mjml>
   ```

2. Render it to HTML in version 4.4.1
3. Send the HTML to Outlook (I used Outlook 2007 on Windows 7)
4. You see pink background (of the wrapper) instead of a kitten

   ![image](https://user-images.githubusercontent.com/11633481/62639881-b52a3b00-b940-11e9-9bfd-be4e97d94676.png)

5. Delete the _z-index_ from compiled HTML code:
   
   ```html
   <!--[if mso | IE]><table role="presentation" border="0" cellpadding="0" cellspacing="0"> <tr><td class="" width="600px" ><table align="center" border="0" cellpadding="0" cellspacing="0" role="presentation" style="width:600px;" width="600" ><tr><td style="line-height:0;font-size:0;mso-line-height-rule:exactly;"><v:image style="border:0;height:300px;mso-position-horizontal:center;position:absolute;top:0;width:600px;z-index:-3;" src="https://placekitten.com/600/300" xmlns:v="urn:schemas-microsoft-com:vml" /><![endif]-->
   ```

6. And it renders the kitten:

   ![image](https://user-images.githubusercontent.com/11633481/62640130-2964de80-b941-11e9-94b3-8e96065ce91b.png)

**MJML environment (please complete the following information):**
- OS: macOS
- MJML Version: 4.4.1
- MJML tool used: NodeJS
- Email sending environment: [Putsmail](https://putsmail.com/)

**Affected email clients (for rendering issues):**
[...]

PR #1706, opened 2019-09-10
add info to doc, related to #1671

kmcb777:


kmcb777:
(the errored build is just an unrelated travis error, works on 2/3 node versions, it's just a doc change anyway)

Issue #1724, opened 2019-10-02
Add letter-spacing attribute to mjml-button


bvonhalle:
**Is your feature request related to a problem? Please describe.**
Our button styles require letter-spacing (they are all caps and it looks much better with the letter-spacing). This seems like a relatively simple, straightforward addition.

**Describe the solution you'd like**
Just add `letter-spacing` as an attribute.

**Describe alternatives you've considered**
Allow for mjml-text or a span inside a button as a hack.

**Additional context**

HenriqueFadoni:
Is this issue still open? I want to solve it

ngarnier:
Hi @HenriqueFadoni, thanks but you can see above that a PR was submitted and merged, so it'll be closed in the next release.

bvonhalle:
Thank you all, this was my first open-source issue and it was solved within 10 days, so cool!

This community gives me faith in humanity.

PR #1732, opened 2019-10-12
Adds letter spacing to mjButton

milap1296:
Adds an attribute letter-spacing to mjButton tag. #1724

**Problem:**
Mj Button didn't have anything such as letter spacing and it would become tidy when button contains only CAPS characters.
Screenshot 2019-10-13 at 4 43 01 AM

**Expected Behaviour**
It should instead look very neat even with just 2px letter spacing.
Screenshot 2019-10-13 at 4 43 35 AM


iRyusa:
👍 Thanks for the pr

PR #1776, opened 2019-12-17
Set bgcolor on mj-group table

btd:
This is to fix outlook 2016 bug. I have no idea why but it just ignore background-color at top div.


iRyusa:
Looks good to me, need to try it on EOA to see if everything is fine 👍 for the PR

btd:
@iRyusa thanks, i added small check for background color 'none' also.

woodymendoza:
Thank you I was looking for a solution for this.  Hoping this gets merged soon.

PR #1783, opened 2019-12-26
mj-include not including files without .mjml in the config string

davidsneighbour:
Fixes mjml/mjml#1763


iRyusa:
We may need to get back the old behavior instead of this one

iRyusa:
Merging from now to reflect the proper behavior

PR #1788, opened 2020-01-07
[add] usemap attribute to image tag

vinikira:
Added [usemap](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/map) attribute to image tag.
It can be used with `map` tag inside of the `mj-raw` component.


iRyusa:
We forgot about this one @vinikira can you fix those linting issue please so we can merge this ?

vinikira:
Done!

iRyusa:
And forgot to mention about the documentation, can you add it to readme.md in mj-image ?
Maybe you should mention that attribute isn't supported everywhere

PR #1820, opened 2020-02-15
Fix_1819; remove closing tags for HTML <img> tags; documentation only

GarryFlemings:
# Included Here #

* A copy of `packages\mjml-image\README.md`, different only by changing

> It is similar to the &lt;img /&gt; tag.

to

> It is similar to the HTML &lt;img&gt; tag.

* Changes to some instances of the HTML code we execute in the documentation for modern browsers. I deleted the closing tags on `` tags.

* Some addition of `alt` attributes on `` tags to help accessibility. Some of this might conflict with other such work in other commits. 😔 Pick the ones you prefer.

* No changes to example code in our documentation (visible to our readers). No closing tags found.

* _No changes to any JavaScript file._


iRyusa:
Yup we have some conflict here :) This is not a big issue where both are accepted. I think we can close this one it's not really an issue

PR #1821, opened 2020-02-15
Fix 1769; documentation only

### LINKED ISSUE #1769: Docs say `font-weight` of `mj-text` can only be a `number` while it also can be a `string`
**Bug description:**
So it really is just a mistake in the docs or maybe an unintended behaviour of the code, but the docs of `mj-text` say `font-weight` is supposed to be a `number` while it actually can also be a `string` like "bold" for example just like in the browser. So I guess the docs need to be adjusted to reflect that behaviour.

**Steps to reproduce:**
1. Create a file with this MJML code: `
<mjml>
  <mj-body>
    <mj-section>
      <mj-column>
        <mj-text font-weight="bold">
          Bold text
        </mj-text>
      </mj-column>
    </mj-section>
  </mj-body>
</mjml>
`
2. Preview the result
3. Text is bold

**MJML environment:**
 - OS: MacOS 10.15.1
 - MJML Version: 4.5.1
 - MJML tool used: MJML App

### LINKED ISSUE #1825: Mismatches between (1) MJML component "index.js" files (objects "allowedAttributes" and "defaultAttributes") and (2) MJML documentation.
**Describe the bug**

This report extends Issue #1769, which reported a difference between `mj-text` code performance and documentation. PR #1821 solves the problem reported (and others) for `mj-text`.

Our documentation would be best if attribute tables for each MJML component matched the corresponding `index.js` files and within them, objects `allowedAttributes` and `defaultAttributes`.

**To Reproduce**

Compare attribute tables in MJML documentation to objects `allowedAttributes` and `defaultAttributes` in corresponding `index.js` files.

**Expected behavior**

Units (column 2 of attribute tables) match that information in `allowedAttributes`. Defaults (column 4) match that information in `defaultAttributes`.

**Proposed Life of this Issue**

The first PR working toward solving this issue is #1821, proposed to fix `mj-text`. That PR is _not_ fix-1825.

Possibly, we'll use a group of PRs to solve all parts of the issue. This issue might close when work is complete for all MJML components.

GarryFlemings:
# Fixes #

This PR fixes #1769.

It is necessary, but not sufficient, to fix #1825. It fixes `mj-text` for #1825. 

# Included Here #

Fixes for the documentation of the `mjml-text` component (file `packages/mjml-text/README.md`), both the one cited in Issue #1769 and other mismatches.

Additional wording about attributes in files `doc/components_1.md` and `doc/components_2.md`.

*No changes to any JavaScript file.*

# Support #

[examples of support for values outside the documentation](https://mjml.io/try-it-live/S1eUg0Pg8)

[example of MJML code supporting same-name attributes for MJML and HTML](https://mjml.io/try-it-live/H1LbbsI1I)

# Unresolved anchor #

My IDE reports it cannot resolve a reference in the mj-text attributes table. On line _css-class_, I refer to _#mjml-style_, which I intend to refer to that other place in documentation. If I haven't found the right syntax, I'll appreciate help with that.


doc/components_1.md
```diff
@@ -27,15 +28,97 @@ For instance, the `mj-button` component is, on the inside, a complex HTML layout
...
+Sometimes, it indicates equivalence with the same-name HTML attribute. 
+
+The _default value_ column specifies the default value for the attribute.
+Attributes without a default value have _n/a_ here.
+
+<!--- Not ambiguous because MJML does not support "n/a" as any attribute value. --> 
+
+<aside class="notice">
+MJML syntax for <em>HTML</em> attributes differs from syntax for <em>MJML</em> attributes.
+MJML "css-class" attributes introduce HTML attributes in MJML components;
```
iRyusa: Not sure about this one, css-class is a global body attribute that refer to `class` HTML attribute applied to the "top" level of HTML generated by a component. 
But you can style mention that you can target those with mj-style

doc/components_1.md
```diff
@@ -27,15 +28,97 @@ For instance, the `mj-button` component is, on the inside, a complex HTML layout
...
+  with the same spelling and different values.
+In such code, MJML uses the HTML attribute
+  inside an MJML "css-class" attribute and
+  in raw HTML inside a "class" attribute.
+It uses the MJML attribute otherwise.
+</aside>
+
+<!--- https://mjml.io/try-it-live/H1LbbsI1I -->
+
+<aside class="notice">
```
iRyusa: Yeah not sure about this paragraph, it's a bit weird to say this where we try to solve that issue with MJML.

We could add a list of resource to check css compatibility like https://www.caniemail.com/

doc/components_1.md
```diff
@@ -27,15 +28,97 @@ For instance, the `mj-button` component is, on the inside, a complex HTML layout
...
+Both have the same purpose of `<head>` and `<body>` in an HTML document.
 
 attribute | unit | description | default value
 ----------|------|-------------|---------------
-owa | string | if set to "desktop", will force desktop display on outlook. Otherwise outlook will display mobile version since it ignores media queries | none
-lang | string | used as `<html lang="">` attribute | none
+owa | string | If set to "desktop", will force desktop display on outlook. Otherwise outlook will display mobile version since it ignores media queries | n/a
+lang | string | same as HTML's `<html lang="">` | n/a
 
+<aside class="notice">
```
iRyusa: Here I don't think this is necessary the description part in the md table is enough

doc/components_2.md
```diff
@@ -24,8 +24,8 @@ tags to make it easier to preview outside the main template
 </mjml>
 ```
 
-The `MJML` engine will then replace your included files before starting the rendering process
+The `MJML` engine will then replace your included files before starting the rendering process.
 
 <aside class="notice">
```
iRyusa: This is not true anymore so we can remove it

packages/mjml-text/README.md
```diff
@@ -1,6 +1,15 @@
 ## mjml-text
 
-This tag allows you to display text in your email.
+Most important, this element places text in your email.
```
iRyusa: I'm not sure about those 2 paragraphs too, I feel like we should mention that mj-text can contain any HTML you want but we expect from MJML developers to already know about HTML so the `p`/ `br` tag part is a bit overkill.

iRyusa: And mj-text example alreay have an example with h1 too :)

packages/mjml-text/README.md
```diff
@@ -19,32 +30,36 @@ This tag allows you to display text in your email.
...
-</p>
+</p> 
 
 <aside class="notice">
-  `MjText` can contain any HTML tag with any attributes. Don't forget to encode special characters to avoid unexpected behaviour from MJML's parser
+Don't forget to encode special characters to avoid unexpected
+  behaviour from MJML's parser.
 </aside>
 
- attribute                    | unit          | description                                 | default value
```
iRyusa: I have a small concern about this too, if we change the way we describe unit we should modify this for all the docs and harmonize everything.

packages/mjml-text/README.md
```diff
@@ -19,32 +30,36 @@ This tag allows you to display text in your email.
...
+font-size      | px: not negative      | text size <!--- not CSS "font-size" because different shape -->      | 13px
+font-style        | often "normal", "italic", or "oblique &lt;angle&gt;"      | same as CSS "font-style"      | n/a
+font-weight                   | string            | font weight (or boldness); same as CSS "font-weight"      | n/a
+height      | px, %      | height of the "mj-text" element; % is relative to the containing block's height. <!--- not CSS "height" because different shape -->      | n/a
+letter-spacing       | px: pos, zero, or neg      | letter spacing within the line <!--- not CSS "letter-spacing" because different shape -->      | n/a
+line-height          | px, %       | space between lines; % is relative to the block's font <!--- not CSS "line-height" because different shape -->      | 100%
+padding              | px, % {1,4}      | sets "padding-bottom", "-left", "-right", and "-top"; % is relative to "width" of containing block <!--- not CSS "padding" because different shape -->      | 10px 25px
+padding-bottom       | px, %                            | sets bottom padding; % same as above        | 10px
```
iRyusa: Padding are not supported in % in some desktop client, I think they are described as px only in the code ?

packages/mjml-text/README.md
iRyusa: As same for HTML, we expect MJML developer to be aware of CSS attribute. So I don't think we should be as verbose here

doc/components_2.md
GarryFlemings: Done.

packages/mjml-text/README.md
GarryFlemings: Fixed.

packages/mjml-text/README.md
GarryFlemings: Shortened.

packages/mjml-text/README.md
GarryFlemings: Should we match `packages/mjml-text/src/index.js > allowedAttributes`? Lines 20-24, part of `allowedAttributes` are

>'padding-bottom': 'unit(px,%)',
    'padding-left': 'unit(px,%)',
    'padding-right': 'unit(px,%)',
    'padding-top': 'unit(px,%)',
    padding: 'unit(px,%){1,4}',

doc/components_1.md
GarryFlemings: I think I understand and agree with everything in the comment.

Not sure what to change. Sorry.

doc/components_1.md
GarryFlemings: Fair. I'll make the case for leaving this in, but I'm happy to remove it. I certainly agree MJML protects the developers from these problems in many cases (hopefully most or almost all).

That said, we know Outlook doesn't support padding well. Or media queries. GMail doesn't support fonts. And ... you know far better than I.

I'll add the caniemail.com reference. And plug MJML as helping handle some of the problems. Maybe that'll add appropriately.

doc/components_1.md
GarryFlemings: I'll change as you like. Maybe I only made minimal changes here to the existing? Dunno what to change.

packages/mjml-text/README.md
iRyusa: ok my bad looks like we've put px  & padding on everything let's keep it that way

doc/components_1.md
```diff
@@ -27,16 +27,102 @@ For instance, the `mj-button` component is, on the inside, a complex HTML layout
...
+The _attribute_ column is a complete list of attributes a component supports.
+An examples of syntax for using a attributes appears just below here
+  in discussion of the `mjml` tag and its `lang` attribute.
+
+<aside class="notice">
+In general, if MJML has an attribute of the same name as an HTML attribute,
+  the two operate similarly.
+</aside>
+
+<!--- NOTE: I understand Markdown recognizes this three-dash comment
```
iRyusa: We should remove comments in doc too imo

doc/components_1.md
```diff
@@ -27,16 +27,102 @@ For instance, the `mj-button` component is, on the inside, a complex HTML layout
...
+MJML syntax for <em>HTML</em> attributes differs from syntax for <em>MJML</em> attributes.
+MJML "css-class" attributes introduce HTML attributes in MJML components;
+  they refer to [mj-style](#mj-style) definitions.
+MJML attributes use the same syntax in MJML tags as HTML attributes in HTML tags.
+</aside>
+
+<aside class="notice">
+MJML can support both an HTML attribute and an MJML attribute
+  with the same spelling and different values.
+In such code, MJML uses the HTML attribute
```
iRyusa: I'm not sure what you want to describe here ?

doc/components_1.md
iRyusa: The whole aside block imo isn't really relevant as you have multiple example of MJML in the doc

doc/components_2.md
```diff
@@ -24,8 +24,6 @@ tags to make it easier to preview outside the main template
 </mjml>
 ```
 
-The `MJML` engine will then replace your included files before starting the rendering process
-
 <aside class="notice">
-Note that the file must be a file with a `.mjml` extension
+The file must have an '.mjml' extension.
```
iRyusa: Whole line can be removed as it's no longer the case with current mj-include implem

packages/mjml-text/README.md
```diff
@@ -1,6 +1,11 @@
 ## mjml-text
 
-This tag allows you to display text in your email.
+Most important, this element places text in your email.
+
+Also important, this element supports placement of raw HTML tags
+  and their attributes anywhere within.
+Don't forget to encode special characters to avoid unexpected
```
iRyusa: This was an issue in MJML3, and it's no longer the case so we can remove this phrase

doc/components_1.md
GarryFlemings: All instances of `<!---` and related text deleted.

packages/mjml-text/README.md
```diff
@@ -1,6 +1,15 @@
...
+Most important, this element places text in your email.
+
+Also important, this element supports placement of raw HTML tags
+  and their attributes anywhere within.
+Sometimes, developers put _only_ HTML within a given `<mj-text>` tag set.
+
+Inside this tag, there's no need to surround each paragraph with HTML `<p>` tags.
+The HTML `<br/>` tag is convenient for creating  line breaks.
+
+
```
GarryFlemings: An email seemed to indicate you wanted me to remove, "Don't forget to encode special characters to avoid unexpected behaviour from MJML's parser". ("This was an issue in MJML3, and it's no longer the case so we can remove this phrase").

I removed it. I mention it because since your note isn't reproduced here, I could have misinterpreted.

doc/components_2.md
iRyusa: Yeah current mj include can include any file don't think we have to describe every behavior, we should describe uncommon cases let's remove the whole aside here

packages/mjml-text/README.md
iRyusa: Let's revert this one for this PR, we can still discuss on slack about this one for a future improve but as stated the linter will provide accurate enough results for now

doc/components_1.md
iRyusa: Don't think this is an edge case at all. It would behave as expected on HTML here, and it will behave perfectly the same in HTML.

doc/components_1.md
iRyusa: I think a simple warning that you should still check if your CSS attributes/values are supported everywhere before using them and you can check with caniemail.

doc/components_1.md
GarryFlemings: This wording is now lines 65-73. I tweaked it a bit. Maybe it's easier to read, now.

iRyusa: I'm still not sure where you want to go with that paragraph, it works here as expected in any HTML/CSS document so why would we need to describe this ?

doc/components_1.md
GarryFlemings: Lines 75-78.

doc/components_1.md
```diff
@@ -27,16 +27,78 @@ For instance, the `mj-button` component is, on the inside, a complex HTML layout
 </table>
 ```
 
+#### Attribute lists
```
iRyusa: I have still mixed feelings about this whole section, here you describe how a table works and as reading this i'm feeling a bit infantilized.

doc/components_2.md
GarryFlemings: Gone.

packages/mjml-text/README.md
GarryFlemings: I'm not confident what part of the file this conversation refers to.

Maybe you want to revert all changes to the attributes table. I'll do it.

Maybe you want me to streamline the table. I tried. Lines 34-53.

iRyusa: As said if we want to tweak those descriptions like that, then we need to do the same for everywhere.

doc/components_1.md
GarryFlemings: "infantilized": Great word, but never in a million years would I have come up with it.

On the other hand, I feel infantilized about the mismatch of numbers here to my screen. I "pulled next" and everything got clearer. 😔

I see your point. I shortened it a lot. Good feedback.

doc/components_1.md
GarryFlemings: My bad. 🤦‍♂️ I should have used the word "class", not "attribute".

And while the content here (with the substitution) might be useful for our readers, it's hard to say it belongs under a header "Attribute lists". Sigh.

Lots of changes.

packages/mjml-text/README.md
GarryFlemings: "Don't forget to encode special characters ...": gone

doc/components_1.md
GarryFlemings: Took another look. Shortened and tweaked.

doc/components_1.md
GarryFlemings: Well, I thought at one point that this web page was updating line numbers for new commits. Maybe not.

Maybe this web page is now telling me the offending `aside` is at line 38. I'm unable to find any commit, let alone in this time frame with an `aside` there.

The nearest `aside` I find is the one with opening text 
>MJML syntax for <em>HTML</em> attributes differs from syntax for ...

We've deleted that one.

PR #1827, opened 2020-02-17
mj-social: Allow border-radius to be percentage or px based

zslabs:
This matches the Button component https://github.com/mjmlio/mjml/blob/master/packages/mjml-button/src/index.js#L13


packages/mjml-social/src/Social.js
```diff
@@ -3,7 +3,7 @@ import { BodyComponent } from 'mjml-core'
 export default class MjSocial extends BodyComponent {
   static allowedAttributes = {
     align: 'enum(left,right,center)',
-    'border-radius': 'unit(px)',
+    'border-radius': 'string',
```
iRyusa: unit(px,%) would be better here ?

zslabs: 🤦‍♂ Yep, you're 100% right - updated. Thanks!

Issue #1814, opened 2020-02-11
I got background-position and background-size working for Outlook for desktop 2007+


Alexnortung:
# Is your feature request related to a problem? Please describe.
*A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]*

When adding a background image to a section there is currently no way to change the background position straight from MJML to work for all clients. I get frustrated thjat I need to make this work by manipulating the HTML and VML.

**I am only describing how to do this with VML objects as it is trivial to make it work with a simple css property**

# Describe the solution you'd like
*A clear and concise description of what you want to happen.*

I would like to be able to add the background-position attribute to the `mj-section` element like this

 ```html
<mj-section 
    background-url="https://placehold.it/900" 
    background-size="cover" 
    background-position="center">
    <!-- content -->
</mj-section>
```

To make the background-position work for Outlook desktop clients the VML should be changed, more specifically, the `<v:fill>` element

## HTML and VML for center background image

To make the image stay in the center the `origin` and `position` attributes needs to be changed to `0.5, 0.5` 
for example

```html
<v:fill  origin="0.5, 0.5" position="0.5, 0.5" src="http://placehold.it/1800x1800" type="tile" />
```

## making cover work

cover will work by adding two more attributes to the `<v:fill>` element: `size="1,1"` and `aspect="atleast"`.
For an example we can use the above example where both cover and position center will work

```html
<v:fill  origin="0.5, 0.5" position="0.5, 0.5" src="http://placehold.it/1800x1800" type="tile" size="1,1" aspect="atleast" />
```

## making contain work 

This one is a little different, there are a few things that should be changed here. The first things that should be changed is `type="frame` this will also make the background non-repeating (which might be a plus 🤠). you also want to add `aspect="atmost"` and `size="1,1"`. 
The `origin` and `position` attributes also needs to be changed to `0,0` to center the image.

For example

```html
<v:fill  origin="0, 0" position="0, 0" src="http://placehold.it/1800x1800" type="frame" aspect="atmost" size="1,1" />
```

# Additional context
*Add any other context or screenshots about the feature request here.*

All of the above examples has been tested in litmus for Outlook versions 2007, 2010, 2013, 2016 and 2019 for windows desktop.

iRyusa:
This could be great to implement, last client to not support something similar would be outlook.com

Can you provide a litmus link for the test for all values ?

Alexnortung:
[...]

PR #1831, opened 2020-02-19
Background-position and background-size working in Outlook for windows desktop and yahoo

Alexnortung:
Based on: https://github.com/mjmlio/mjml/issues/1814

Background position and background size is now working with yahoo and outlook desktop.
I have only added `background-posiiton-x` and `background-position-y`, as it seemed to complicated to implement just `background-position`.
`background-repeat: no-repeat` will now also work for outlook, the only issue is that the image will be stretched if `background-size` is not set to either `cover` or `contain`.

Tests:
https://litmus.com/builder/9ebafc2
https://litmus.com/builder/fa7f3af
https://litmus.com/builder/06bf715
https://litmus.com/builder/2f46c4e
https://litmus.com/builder/ed180d2
https://litmus.com/builder/3fb5dbf

Edit:
I have added support for percentage values instead of just top, center, bottom, left and right.
Here are my tests for that
https://litmus.com/builder/4cdafaa
https://litmus.com/builder/ab592f7

### DIFF
--- a/packages/mjml-section/src/index.js
+++ b/packages/mjml-section/src/index.js
@@ -8,6 +8,8 @@ export default class MjSection extends BodyComponent {
     'background-url': 'string',
     'background-repeat': 'enum(repeat,no-repeat)',
     'background-size': 'string',
+    'background-position-x': 'enum(left,center,right)',
+    'background-position-y': 'enum(top,center,bottom)',
     border: 'string',
     'border-bottom': 'string',
     'border-left': 'string',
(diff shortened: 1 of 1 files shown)


iRyusa:
We won't be able to ship it with 4.6 as it requires some testing and "might" change a bit the current behavior. That being say let's target this for a 4.7 / 5.0 with custom attributes first implementation

iRyusa:
Ok we should take a look to get this merge cc @kmcb777

kmcb777:
very nice improvement thanks

ezramechaber:
This is a wonderful improvement and solved a major issue I was having. Thank you!

PR #1835, opened 2020-02-20
fix-1834, documentation only

GarryFlemings:
Included: A new "packages\mjml-column\README.me" different only by adding a cautionary note about column widths in a section summing over 100%.

No changes to any JavaScript file.


iRyusa:
I think you've missed some changes on your commit

GarryFlemings:
> I think you've missed some changes on your commit


erf ... yup. I have changes ready to send depending on your answer to the Slack question.

PR #1838, opened 2020-02-25
Setup yarn workspaces

TrySound:
Yarn workspaces simplify upgrading dependencies, for example with
`yarn upgrade-interactive --latest` command.

Also easier to track what's in node_modules.


iRyusa:
Had a quick review this seems nice, I think we can merge that after 4.6 ?

TrySound:
I wanted to submit a couple of more PRs. For example with migrating to babel 7 without core-js dependency.

iRyusa:
That would be great ! Feel free to ping me on slack if you need anything

TrySound:
@iRyusa Looks like errors came from other committer.
image

iRyusa:
Yup @kmcb777 will fix it soon sorry about that

TrySound:
@iRyusa I rebased. Looks like no-unresolved lint is fixed with yarn workspaces. Good to merge?

PR #1840, opened 2020-02-27
Adding "inner-background-color" to mj-column to improve gutter

iRyusa:


iRyusa:
Let's add inner-border too @kmcb777

PR #1861, opened 2020-03-10
Upgrade to babel v7

TrySound:
Added `rimraf` and `@babel/cli` to each package to prevent using paths
to run scripts. In yarn 2 it is forbidden because no node_modules there.


DRoet:
nice work!

side note: @TrySound I noticed you have some experience w/ rollup. Do you think it is technically possible to strip out the nodejs specific stuff (minify/mj-include) to provide a slightly leaner browser-only build (less functionality) using rollup? I've been trying to experiment with this idea but idk if it is feasible.

package.json
```diff
@@ -12,17 +12,13 @@
...
     "babel-eslint": "^7.2.3",
-    "babel-plugin-add-module-exports": "^0.2.1",
-    "babel-plugin-transform-async-to-generator": "^6.24.1",
-    "babel-plugin-transform-decorators-legacy": "^1.3.4",
-    "babel-plugin-transform-regenerator": "^6.26.0",
-    "babel-plugin-transform-runtime": "^6.23.0",
-    "babel-preset-env": "^1.6.0",
-    "babel-preset-stage-0": "^6.24.1",
-    "babel-register": "^6.26.0",
+    "babel-plugin-add-module-exports": "^1.0.2",
```
DRoet: is this one still needed?

TrySound: Yes, to not break commonjs users. Can be removed on major bump.

TrySound:
@DRoet sure, rollup is always a good option. Even for node.js apps. I used rollup to bundle my app backend recently.

package.json
DRoet: whoops I read over it in `babel.config.js`

kmcb777:
Nice work, thanks !

PR #1863, opened 2020-03-13
fix #200 custom attributes :sparkles:

### LINKED ISSUE #200: Ability to add custom properties

kmcb777:


iRyusa:
We have to add a bit of test to ensure custom templating in mj-raw won't break when selectors are applied ?

PR #1877, opened 2020-04-06
Background images size

kmcb777:


kmcb777:
@Alexnortung Hi, during the many tests i made on your PR, i noticed that we could also handle size in px/%, so i added this.
Also, i had to keep the type to "tile" when size is not provided because it's important for us that existing templates will still render the same.
Since you likely made a lot of tests about vml, it would be nice to have your review ;) 
I also added support for background-position attribute (2 values max) because we already do this kind of parsing for other attributes.

PR #1878, opened 2020-04-06
Stick with single lodash import style

TrySound:
With `import _ from 'lodash'` or `import { omit } from 'lodash'` node
requires big module different from modules in `import omit from 'lodash/omit'`.

In this diff I converted all lodash imports to single style to reduce
overhead in required modules.


TrySound:
Treeshaking is processing `import { omit } from 'lodash'` form. But in case of lodash it does not work because of many top level expressions which are considered as side effects. This is why direct import is the recommended way. Same with node. It has to load full lodash.

Yeah, babel plugin may work. But do you think the task worth a plugin?

Btw many lodash functions are supported natively. Maybe remove lodash where possible? I can do this in separate PR.

iRyusa:
Well lodash method tend to be safer when accessing null/undefined structure. As we don't use typescript for now and current test coverage are pretty bad, I think we should stick to lodash. 

It's mostly a matter of readability here, if we can group the imports and having the optimization I think the babel plugin option is fine ? Don't think in a near future we'll ditch babel, so adding an extra plugin is fine.

And it's maybe the safer option as future PR can miss this change too, what do you think ?

iRyusa:
I think we'll close that one, I would prefer to use the babel plugin instead, so we have less things to focus on next CR.

PR #1879, opened 2020-04-08
Reset custom components registered

aitor-hernandez:
It would be interesting to be able to reset the registered components, in order to associate different component configurations.
As far as I could see, MJML App being a single instance once it loads a component keeps it registered without the possibility of deleting it. When a new MJML template is opened all the loaded components are used to render it, it would be interesting to add the possibility that this does not happen.


kmcb777:
Hi @aitor-hernandez, yes we thought about adding an option for this
The issue is that we don't just need to require the component again, node's `require` has a cache and we would need to delete it manually. We'll need to allow this in mjml, then in the app.
But I don't see why you opened a pull request between next and master for this, it should be an issue instead, so i'll close this one (i think there is already an issue for this on the mjml-app repo).

PR #1887, opened 2020-04-28
missing documentation

### LINKED ISSUE #1886: Expose ignoreIncludes as an option for mjml2html
**Is your feature request related to a problem? Please describe.**
ignoreIncludes is an option of MJMLParser, but you can't pass in that option when using mjml's default mjml2html export. For my node.js project, I would like to turn off includes for all my templates easily instead of going through all them and manually removing mj-include statements.

**Describe the solution you'd like**
Accept ignoreIncludes as an option to mjml2html and pass it along to be used when calling MJMLParser.
```
import mjml2html from 'mjml'
const output = mjml2html(`
<mjml>
  <mj-body>
    <mj-include path="./header" /> <!-- or 'header.mjml' -->
  </mj-body>
</mjml>
`, {
    ignoreIncludes: true
});

```

**Describe alternatives you've considered**
I've considered importing MJMLParser myself and passing the output to mjml2html myself, but I would be worried that the interface could change on me without being aware since MJMLParser is not part of the official documentation.

**Additional context**
I'm using mjml v4.6.2

kmcb777:
and remove unsupported attribute on mjml-social-element


packages/mjml-cli/README.md
```diff
@@ -61,13 +61,14 @@ mjml -s input.mjml
 mjml --stdout input.mjml
 ```
 
-### Minify and beautify the output HTML
+### Minify and beautify the output HTML (or use any option available on mjml2html)
```
iRyusa, line 64: Mayeb rename the whole part then to Available options ? Or something like that

PR #1904, opened 2020-05-14
fix wrapper docs for background-url

dmgawel:
background-image property was mistakenly used instead of background-url when describing Outlook issue.


iRyusa:
I think it should be `background image` instead of `background-image` it makes a bit more sense than an url what do you think ?

iRyusa:
Nvm i'll merge that thanks !

PR #1907, opened 2020-05-16
feat: update eslint and prettier

DRoet:
Pretty big diff because of all the rule changes, but should be good to go.

No Eslint 7.x yet because the plugins are not supported yet.

### DIFF
--- a/packages/mjml-section/src/index.js
+++ b/packages/mjml-section/src/index.js
@@ -226,7 +226,7 @@ export default class MjSection extends BodyComponent {
         <tr>
       <![endif]-->
       ${this.renderChildren(children, {
-        renderer: component =>
+        renderer: (component) =>
           component.constructor.isRawElement()
             ? component.render()
             : `
@@ -260,7 +260,7 @@ export default class MjSection extends BodyComponent {
 
     const { containerWidth } = this.context
 
-    const isPercentage = str => /^\d+(\.\d+)?%$/.test(str)
+    const isPercentage = (str) => /^\d+(\.\d+)?%$/.test(str)
 
     let vSizeAttributes = {}
     let { posX: bgPosX, posY: bgPosY } = this.getBackgroundPosition()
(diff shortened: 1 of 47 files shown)


DRoet:
hmm prettier 2.x fails on the node 8 CI but that is because it requires node >= 10, but technically it isn't a breaking change since it is just a devDepencendy.

shouldn't the CI be run w/ `--prod` to avoid this problem?

iRyusa:
I think we should just remove node 8 and move on node 12 as LTS

iRyusa:
Same conflict here with yarn lock 👍

TrySound:
If you are not gonna test features on node 8 you can break some users. Are you gonna start working on v5?

DRoet:
I'll resolve the conflicts after the CI changes are merged

DRoet:
@TrySound yeah I was wondering the same thing, but it seems like the current version strategy isn't truly semver. For example node 6 got dropped around the `4.4.0` release and it's not clear which node versions are directly supported (no engines field in package.json etc.)

iRyusa:
> If you are not gonna test features on node 8 you can break some users. Are you gonna start working on v5?

I don't know if we have to push a major v5 for that, as node 8 no longer supported anymore. Travis.yml were not updated to reflect current node LTS. 4.6.X might be the latest version to officialy support it and then we're moving to 4.7.X branch now.

DRoet:
@iRyusa rebased

packages/mjml-section/src/index.js
```diff
@@ -299,35 +299,38 @@ export default class MjSection extends BodyComponent {
...
+          } else {
+            pos = (-50 + decimal * 100) / 100
+            origin = (-50 + decimal * 100) / 100
+          }
+        } else if (bgRepeat) {
+          // top (y) or center (x)
+          origin = isX ? '0.5' : '0'
+          pos = isX ? '0.5' : '0'
         } else {
-          pos = (-50 + decimal * 100) / 100
```
iRyusa, line 317: Mh i'm not sure about this file here ? it seems that the behavior changed ?

DRoet: @iRyusa yeah idk why the diff is weird, but it is only a indent change. its a lot more clear in a side-by-side view

PR #1909, opened 2020-05-17
feat: drop node 8 and use github actions

DRoet:
see https://github.com/mjmlio/mjml/pull/1907#issuecomment-629800846

Thought it would be interesting to use the builtin github actions instead since it is pretty neat, but I can revert that change if needed.

successful run on my fork: https://github.com/DRoet/mjml/actions/runs/107366295
successful run with 14.x included: https://github.com/DRoet/mjml/actions/runs/110149142


DRoet:
@iRyusa I added node 14.x aswell since it will become the next LTS in october

DRoet:
OK this should be ready to be merged now

iRyusa:
Probably need to remove the travis badge then on readme ?

DRoet:
@iRyusa done

iRyusa:
Looks like we have a yarn lock conflict then we could merge that I think, we'll monitor if the free tier isn't enough then we'll rollback to travis I guess

DRoet:
@iRyusa fixed.

GitHub Actions usage is free for public repositories, I don't believe there are any limits to it currently, but I might be unaware.

DRoet:
@iRyusa After landing https://github.com/mjmlio/mjml/commit/5250617520c790252202a14d77fa53bb28f15e52

The build in this PR actually runs, CI is still green 😃

PR #1912, opened 2020-05-18
Wrong default value for padding on mj-navbar-link

kevingermain:


iRyusa:
👍  thanks you for catching that

PR #1916, opened 2020-05-24
Feat: add sizes and srcset attributs to mj-social-element and mj-image

rickytranmer:
EOA results: https://app.emailonacid.com/app/acidtest/Chhp7sFl0OQRzgB79PRjfDsv2mnm0pET9JZI1isNsuSzH/list
test-mjml
index-html


iRyusa:
Maybe we should add `sizes` attribute on image & social too ?

rickytranmer:
> Maybe we should add `sizes` attribute on image & social too ?

S

rickytranmer:
> Maybe we should add `sizes` attribute on image & social too ?

Sorry, I'm not sure what you mean by this.

iRyusa:
Well people more likely to use srcset with sizes attribute https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images

rickytranmer:
Yes, I could definitely see that as another feature.  My usage is just for retina images, like with mj-image.  Since mj-image didn't have the `sizes` attribute, I didn't add it to mj-social-element.  Should adding `srcset` to mj-social-element be in the same PR as adding `sizes` to both, or split PR?

iRyusa:
Maybe we can add everything in this PR that would be great !

PR #1918, opened 2020-05-25
Automatically normalize lodash imports

TrySound:
Lodash is very big project. To not import everything we have to use
direct imports like `lodash/map`. Though it's hard to maintain such
imports in a big project. We need either linting or automatical
conversion to the right format.

In this diff I added babel plugin to solve the problem and converted all
direct imports to named ones for consistency.


iRyusa:
👍

PR #1923, opened 2020-05-27
fix: resolve some merge conflicts and eslint errors

DRoet:
some cleanup regarding the new babel structure after the `4.7.0` merge that was causing the build to fail + eslint fixes.

@kmcb777


kmcb777:
thanks for this 👍

PR #1932, opened 2020-06-02
Enables the mjml renderer to receive a userContext object that is available to all components on rendering

renanccastro:
**Enables the mjml renderer to receive a userContext object that is available to all components on rendering**

We have some components that depend on user data when transforming the MJML code to HTML. Until now, we were using a global variable for context related things, but this can get complicated as we scale up the number of mjml2html concurrent executions.

Usage example:
```
const html = mjml2html(mjmlContent, {
        userContext: { data },
}).html;
```

Inside custom components we can fetch it by using:

```
export default class PAuthQrCode extends BodyComponent {
    render() {
       const context = this.userContext;
       const href = context.data.href;

       return this.renderMJML(`
      <mj-image-customized css-class="logo" align="center" width="175px" height="auto" padding="0" src="${href}?qrcode.png" />
    `);

    }
}

```

### DIFF
--- a/packages/mjml-core/src/createComponent.js
+++ b/packages/mjml-core/src/createComponent.js
@@ -35,12 +35,14 @@ class Component {
       context = {},
       props = {},
       globalAttributes = {},
+      userContext,
     } = initialDatas
 
     this.props = {
       ...props,
       children,
       content,
+      userContext,
     }
 
     this.attributes = formatAttributes(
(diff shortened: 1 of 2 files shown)


iRyusa:
Well that would fall into templating territory and we're against that. Please apply templating before or after rendering (you can use new `preprocessors` option for that).

PR #1937, opened 2020-06-07
Avoid mixing commonjs and es modules

TrySound:
This is valid only because babel transpiles es modules into commonjs.
For bundlers (except webpack) and native es modules it's not valid code.


iRyusa:
We had some issue with `.default` exports maybe the migration babel 7 fixed it

TrySound:
The problem was because of `require`. You fixed it by using `add-module-exports` plugin. Babel wraps all imports with interop which prevents similar problem even without this plugin. Though it's still required for public api but can be avoided in major bump.

PR #1941, opened 2020-06-15
Fix for empty attributes on beta 1

iRyusa:


DRoet:
Hmm I just noticed the GH actions are running twice in this PR, maybe the `push` event should be limited to just the master branch

PR #1946, opened 2020-06-20
Upgrade yargs and drop commander

TrySound:
mjml-cli uses yargs for implementing cli while mjml-migrate uses
commander. In this diff I upgraded yargs in cli and migrated migrate to
yargs for consistent tools and less dependencies (though yeah, juice
also uses commander).

Also fixed shebang in migrate. `#!/bin/env node` does not work in my
project.


iRyusa:
I think commander is still catched from another dep, but looks good to me, we'll merge this after 4.7 release

TrySound:
> though yeah, juice also uses commander

TrySound:
Can't this fit 4.7? The change is not that big and fixes the bug with migrate cli.

Issue #1947, opened 2020-06-20
Roadmap v5


TrySound:
- [ ] split mjml and mjml-cli - my team only consume mjml2html function from mjml package and does not use cli which adds a lot of dependencies
- [ ] drop mjml-migrate
- [ ] move html-minifier and js-beautify to cli - both adds a lot of dependencies to mjml package, would be good to let non-cli users add for example prettier and no minifier without bloating node_modules
- [x] support browser targets

iRyusa:
> split mjml and mjml-cli - my team only consume mjml2html function from mjml package and does not use cli which adds a lot of dependencies

I'm not sure about this one, `mjml` package should remain the default package to have the full experience, you can still just target core + every other dependencies manually and register components manually here ?

> move html-minifier and js-beautify to cli

May be we should move that part on its own package and let user requires it with .mjmlconfig ?

TrySound:
Sure, any option would be good

HansUXdev:
Is an official roadmap for V5?

dmgawel:
After separating mjml-cli (with minifier and beatuifier) it should be much easier to support browser targets, which I kindly ask to add for v5 roadmap. In-browser email builders with bundled mjml (instead of using API) are a perfect use-case for that.

rootwork:
Is this why there's both `mjml` and `mjml-cli` on npm?

https://www.npmjs.com/package/mjml
https://www.npmjs.com/package/mjml-cli

That really confused me when I went looking to install it and saw references to "the cli".

They have different dependencies, but the readme for the `mjml-cli` npm package seems to just be a different version of the readme for `mjml`. It uses `mjml` throughout, instead of `mjml-cli` (including for the installation!) -- but if you have `mjml-cli` you'd have to run things like `mjml-cli input.mjml`, not `mjml input.mjml`, right?

Or is the npm `mjml-cli` package just a placeholder for now?

iRyusa:
Forgot to reply @rootwork `mjml` is the "community" package where everything is setup to have the standard lib working (bundling the cli + std components). `mjml` package have an extra binary that register every components before passing everything to `mjml-cli`.

`mjml-cli` rely on `mjml-core` but not on `mjml` package.

CanRau:
proper browser support would be amazing as this should enable easier use in Cloudflare Workers I guess 🤞🏼

dmgawel:
[...]

PR #1950, opened 2020-06-21
Add mjml-preset package

TrySound:
Ref https://github.com/mjmlio/mjml/issues/1947#issuecomment-647006838

The new package includes all components and validations extracted from
main mjml package.

Now mjml-cli can be migrated into mjml package to reduce complexity a
little bit.

mjml-preset has `registerPreset` function which modifies global objects
by calling registerComponent and registerDependencies. Though there
might be a better and less side effectful way to define presets.


TrySound:
I'm not a fan of "Prefer default export" lint rule. Not sure how to fix yet.

iRyusa:
Really like it, it's way easier to use mj core without mj cli with that preset 👍 Can we wait a bit to finish MJML 4.7 before merging this?

DRoet:
any ETA on the 4.7 release? really looking forward to the 5.x stuff that @TrySound is proposing

TrySound:
Btw, maybe better provide declarative structure which is passed to mjml core instead?

DRoet:
@TrySound could you try and give a small example of what you are suggesting?

PR #1952, opened 2020-06-21
Replace .npmignore with files field

TrySound:
Allowlist can prevent from publishing useless files like here

https://unpkg.com/browse/react-frame-component@4.1.2/

See "tags"


iRyusa:
👍  thanks for that

PR #1958, opened 2020-06-28
fix conditional comment for meta tag in skeleton

cheapsteak:
`<!--[if !mso]><!-- -->` will be parsed as a comment with a text content of `[if !mso]><!-- `

This PR changes the conditional comment in the skeleton to matches the other instances of conditional outlook tags in conditionalTag.js 

https://github.com/mjmlio/mjml/blob/604a9f2d2c19a224590814615fda8666fe1fd4e4/packages/mjml-core/src/helpers/conditionalTag.js#L4-L5

### DIFF
--- a/packages/mjml-core/src/helpers/skeleton.js
+++ b/packages/mjml-core/src/helpers/skeleton.js
@@ -30,7 +30,7 @@ export default function skeleton(options) {
         <title>
           ${title}
         </title>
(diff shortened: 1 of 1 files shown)


iRyusa:
@kmcb777 can you test this one ? This seems like a small patch and might fix post process from some ESP

kmcb777:
It doesn't seem to cause any problems 👍

PR #1969, opened 2020-07-07
Fix 1967 docs only

GarryFlemings:
Improves "mj-hero" documentation (mjml-hero/README.md), largely without introducing new subjects. Some material moves from "&lt;aside&gt;" elements to normal text. Some material gets some rephrasing.

No changes to any JavaScript file.


packages/mjml-hero/README.md
```diff
@@ -1,5 +1,23 @@
 ## mj-hero
 
+This element displays an image in the background of an email.
```
iRyusa, line 3:
```suggestion
This element displays an hero image in the background of an email.
```

packages/mjml-hero/README.md
```diff
@@ -80,24 +98,18 @@ Fluid height
...
-</aside>
-
-<aside class="notice">
-For better result we encourage you to use a background image width equal to the hero container width and always specify a fallback background color, in case the user email client does not support background images.
+   The "height" attribute is required only for 'mode="fixed-height"'. 
 </aside>
 
 <aside class="notice">
-  Please keep the hero container height below the image height. When the hero container height - both in fixed or fluid modes - is greater than the background image height, we can’t guarantee a perfect rendering in all supported email clients
+  Outlook.com does not support 'background-position' with
```
iRyusa, line 111: I think it does now, Outlook desktop doesn't but we don't really need to give that info here

packages/mjml-hero/README.md
```diff
@@ -1,5 +1,23 @@
...
+  `background-url`.
+
+In both fixed and fluid modes, it's best to use an image with width
+  the same as the `mj-body` width (`width="600px"` by default).
+Further, it's best to use an image with height the same or larger
+  than the `height` of `mj-hero`.
+Poor rendering frequently results from image sizes outside
+  this guidance.
+
+For the widest great rendering, always specify `background-color`.
```
iRyusa, line 17: I'm not really ok with that, it should be something like

```suggestion
You should add a `background-color` to provide a fallback color if the email client doesn't support background image.
```

packages/mjml-hero/README.md
```diff
@@ -1,5 +1,21 @@
 ## mj-hero
 
+This element displays a hero image in the background of an email.
+It behaves like an `mj-section` with a single `mj-column`.
+
+Set `background-height` and `background-width`
```
iRyusa, line 6:
```suggestion
`background-height` and `background-width` attributes are mandatory
```

packages/mjml-hero/README.md
```diff
@@ -1,5 +1,21 @@
...
+
+In both fixed and fluid modes, it's best to use an image with width
+  the same as the `mj-body` width (`width="600px"` by default).
+Further, it's best to use an image with height the same or larger
+  than the `height` of `mj-hero`.
+Image sizes outside this guidance often result in poor rendering.
+
+Use `background-color` to provide a fallback color
+  in case an email client doesn't support `background-url`.
+
```
iRyusa, line 18: Those use to be warnings with aside

packages/mjml-hero/README.md
```diff
@@ -1,5 +1,21 @@
 ## mj-hero
 
+This element displays a hero image in the background of an email.
+It behaves like an `mj-section` with a single `mj-column`.
+
+Set `background-height` and `background-width`
+  to the actual height and width of the image identified in
+  `background-url`.
+
+In both fixed and fluid modes, it's best to use an image with width
```
iRyusa, line 10: > It's best to use an image with width matching `mj-body`'s width (`width="600px"` by default).

No need to mention mode here it's irrelevant

packages/mjml-hero/README.md
```diff
@@ -1,5 +1,21 @@
...
+  the same as the `mj-body` width (`width="600px"` by default).
+Further, it's best to use an image with height the same or larger
```
iRyusa, line 12: For better results, it's best to use an image with height the same or larger
```suggestion
Further, it's best to use an image with height the same or larger
```

So we can remove `Image sizes outside this guidance often result in poor rendering.`

packages/mjml-hero/README.md
```diff
@@ -80,24 +96,18 @@ Fluid height
...
-</aside>
-
-<aside class="notice">
-For better result we encourage you to use a background image width equal to the hero container width and always specify a fallback background color, in case the user email client does not support background images.
+   The "height" attribute is required only for 'mode="fixed-height"'. 
 </aside>
 
 <aside class="notice">
-  Please keep the hero container height below the image height. When the hero container height - both in fixed or fluid modes - is greater than the background image height, we can’t guarantee a perfect rendering in all supported email clients
+  Email client support of 'background-position' with
```
iRyusa, line 109: Just remove that, it doesn't have any purpose here

packages/mjml-hero/README.md
GarryFlemings: True. I'll say "asides" and sidebars work better when they're supplementing the existing text. We didn't have any. Once I looked at it, I hoped you'd feel this material makes fine text.

packages/mjml-hero/README.md
GarryFlemings: "For better results" is wordy. And tolerable. All good.

packages/mjml-hero/README.md
GarryFlemings: I think your direction to remove is about the css-class line in the attribute table. They're not in the other tables. I removed it from this one, too.

iRyusa: Nope this won't give any information here, you should just remove this aside.

packages/mjml-hero/README.md
```diff
@@ -1,5 +1,21 @@
 ## mj-hero
 
+This element displays a hero image in the background of an email.
+It behaves like an `mj-section` with a single `mj-column`.
+
+`background-height` and `background-width` attributes are mandatory.
+  to the actual height and width of the image identified in
+  `background-url`.
```
iRyusa, line 8: This need to be remove then

GarryFlemings: gone

packages/mjml-hero/README.md
```diff
@@ -1,5 +1,27 @@
 ## mj-hero
 
+This element displays a hero image in the background of an email.
```
iRyusa, line 3:
```suggestion
This element displays a hero image.
```
It would be a bit misleading here.

packages/mjml-hero/README.md
```diff
@@ -1,5 +1,27 @@
...
+  in case an email client doesn't support `background-url`.
+
+<aside class="notice">
+  The "height" attribute is required only for 'mode="fixed-height"'.
+</aside>
+
+<aside class="notice">
+  Email client support of 'background-position' with 'mode="fluid-height"' is
+      not consistent.
+</aside>
```
iRyusa, line 23: This sill need to be removed

PR #1996, opened 2020-08-03
Fix-1995-docs-only; remove MJML3 sentence from docs for mj-text

GarryFlemings:
The docs for mj-text include an &lt;aside> with the sentence, "Don't forget to encode special characters to avoid unexpected behaviour from MJML's parser".

The sentence is a left-over from MJML3. It does not apply to MJML4.

Documentation only. No JavaScript files.


GarryFlemings:
Replaced by PR #2002 (same content, similar title, different branch).
This one's branch is fix-1995-docs-only. Please ignore this branch. Even better: Delete it.

PR #2002 uses branch fix_1995-docs-only (note the underscore in position 4). That's a good one.

PR #2004, opened 2020-08-12
feat: also return the mjml json structure

DRoet:
I know there is the `mjml2json` package out there, but it is a bit outdated (mjml 3) and `mjml-core` basically already does the same thing as far as I can tell.

The only downside perhaps is that this includes properties such as `absoluteFilePath` which we could omit, but I just wanted to gather thoughts on this PR first.

### DIFF
--- a/packages/mjml-core/src/index.js
+++ b/packages/mjml-core/src/index.js
@@ -364,6 +364,7 @@ export default function mjml2html(mjml, options = {}) {
 
   return {
     html: content,
+    json: mjml,
     errors,
   }
 }


iRyusa:
I'm curious about the usage of this ? If you need the json Input won't you just call the parser anyway ?

DRoet:
Well I wanted to save the mjml as json and noticed in the documentation that the `mjml2html()` function already works out of the box with json inputs. Thats when I came across the `mjml2json` package, but noticed it seemed slightly out of date and I started comparing its functionality with the mjml-core package.

I'm sure I could recreate the mjml2json package and run the parser manually, but I thought maybe it would be nice if the `mjml-core` could just return it directly since it is "already there".

iRyusa:
I get that, but what would be the use case of that JSON at the end ?

DRoet:
Oh I'm working on creating a web based mjml builder and having the JSON structure would be a lot easier to work with. Having a way of parsing existing mjml templates into JSON would be very useful.

Issue #1911, opened 2020-05-17
CLI: --validate doesn't use the same base directory


mousedownmike:
**Describe the bug**
When I run `npx mjml -w ./templates/**/*.mjml -o ./build/`, my templates are successfully generated and placed in my build directory.

When I try to validate with the same path `npx mjml --validate ./templates/**/*.mjml`, I get the following errors:
```
Line 7 of /projects/email (mj-raw) — mj-include fails to read file : ../includes/header-logo.mjml at /projects/includes/header-logo.mjml
Line 37 of /projects/email (mj-raw) — mj-include fails to read file : ../includes/footer.mjml at /projects/includes/footer.mjml
```

**To Reproduce**
1. Create a simple project with the directory structure defined below.
2. In the `confirm-sign-up.html` file incorporate the mjml tag: `<mj-include path="../includes/footer.mjml"/>`
3. In the parent directory of `templates` run: `npx mjml -w ./templates/**/*.mjml -o ./build/` and observe that  the included file is properly rendered in the calling template.
4. Again, in the parent directory of `templates` run: `npx mjml --validate ./templates/**/*.mjml` and observe the `mj-include fails to read file` errors.

** Sample Directory Structure**
```
templates/
├── cognito
│   └── confirm-sign-up.mjml
└── includes
    ├── footer.mjml
    └── header.mjml
```

**Expected behavior**
It would seem like the validation command should be able to find the same include files that the mjml command finds when it is watching.

**MJML environment (please complete the following information):**

```
npx envinfo --system --npmPackages
  System:
    OS: Linux 4.15 Ubuntu 18.04.4 LTS (Bionic Beaver)
    CPU: (8) x64 Intel(R) Core(TM) i7-8565U CPU @ 1.80GHz
    Memory: 986.20 MB / 15.32 GB
    Container: Yes
    Shell: 4.4.20 - /bin/bash
  npmPackages:
    mjml: ^4.6.3 => 4.6.3 
```

**Additional context**
I discovered the watch command that I'm using in #177

iRyusa:
Shouldn't be hard to fix this, we forgot to pass down the filePath & actualPath to the parser

PR #2011, opened 2020-08-18
Fix validate command to use the default base directory

YuriGuilherme:
This PR adds the same base directory used in the build command to the validator command.

Close: #1911


iRyusa:
👍 thanks !

PR #2015, opened 2020-08-20
remove extra argument in call to borderParser()

FelixSchwarz:
The function only takes a single argument.


iRyusa:
👍 Was probably a parseInt before

PR #2019, opened 2020-08-28
Upgrade juice

TrySound:
The new juice versions are about upgrading deps mostly.

https://github.com/Automattic/juice/releases/tag/v6.0.0
https://github.com/Automattic/juice/releases/tag/v7.0.0

The positive point is juice 7 is almost two times smaller.
https://packagephobia.com/result?p=juice@7.0.0

Mostly simplified web-resource-inliner
https://github.com/jrit/web-resource-inliner/compare/v4.3.2...v5.0.0

### DIFF
--- a/packages/mjml-core/package.json
+++ b/packages/mjml-core/package.json
@@ -26,7 +26,7 @@
     "cheerio": "1.0.0-rc.3",
     "html-minifier": "^3.5.3",
     "js-beautify": "^1.6.14",
-    "juice": "^5.2.0",
+    "juice": "^7.0.0",
     "lodash": "^4.17.15",
     "mjml-migrate": "4.7.0-beta.2",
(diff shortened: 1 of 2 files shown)


DRoet:
Looks good, but this would mean dropping node 8 on top of dropping node 6, which I would be in favor for since they are both EOL.

TrySound:
Well, they are already dropped
https://github.com/mjmlio/mjml/blob/master/.github/workflows/mjml-workflow.yml#L8

DRoet (approved):

TrySound:
Hey @kmcb777 @iRyusa What do you think?

kmcb777:
Hey @TrySound , thanks for this !
I'll include this in 4.7 right now as there's not much risk of side effects

PR #2033, opened 2020-09-16
Mj include improve wording

GarryFlemings:
There's improvable wording in a previous commit that hasn't yet made it to the Web.
This fixes it.

file: doc/components_2.md
no other; no JavaScript files


doc/components_2.md
```diff
@@ -25,5 +25,5 @@ You can wrap your external mjml files inside the default `mjml > mj-body`
 </mjml>
 ```
 
-The `MJML` engine will then replace your included files before starting
-  the rendering process.
+The `MJML` engine will then include these files
+  in the HTML file as it builds the HTML.
```
iRyusa, line 29: I don't think this is really better as it's replaced way before it "builds" the HTML

GarryFlemings: Good. A fix to that makes the whole thing even better.

I think the biggest point is that the MJML engine doesn't really touch the rendering process. That happens long after MJML completes.

PR #2042, opened 2020-09-24
feat(mj-accordion-text): add new attributes

brunohkbx:
This PR adds the following attributes to the component `mj-accordion-text`

1. font-weight
2. letter-spacing
3. line-height

## Motivation
It would be nice if `mj-accordion-text` accepts the same attributes as `mj-text` does. 
I had to change the `font-weight` from the accordion content and I couldn't because `mj-accordion-text` does not accept it.


iRyusa (commented):
Can you add those line into the readme too please ?

brunohkbx:
@iRyusa done :+1:

iRyusa (approved):

iRyusa:
Thanks !

PR #2046, opened 2020-09-29
renderMJML supports sibling elements

IanEdington:
This PR adds support for returning sibling elements from a custom component. Based on slack conversation: https://mjml.slack.com/archives/C12HESC2G/p1601316349042200

I tried to add the functionality without breaking any of the other code paths.

I tested this with components that return 1, 2, or 8 sibling nodes.
However, the testing framework is difficult to work with and it seems like it would be a LOT of work to actually test this properly.

Let me know if there's anything else this PR needs 🥂

### DIFF
--- a/packages/mjml-core/src/createComponent.js
+++ b/packages/mjml-core/src/createComponent.js
@@ -72,6 +72,20 @@ class Component {
 
   renderMJML(mjml, options = {}) {
     if (typeof mjml === 'string') {
+      if (mjml.indexOf('<mjml>') === -1) {
+        // supports returning siblings elements from a custom component
+        const partialMjml = MJMLParser(`<mjml><mj-body>${mjml}</mj-body></mjml>`, {
+          ...options,
+          components,
+          ignoreIncludes: true,
+        })
+        const body = find(partialMjml.children, { tagName: 'mj-body' })
+
+        return body.children
+          .map(child => this.context.processing(child, this.context))
+          .join('')
+      }
+
       mjml = MJMLParser(mjml, {
         ...options,
         components,
--- a/packages/mjml-core/src/createComponent.js
+++ b/packages/mjml-core/src/createComponent.js
@@ -72,25 +72,15 @@ class Component {
 
   renderMJML(mjml, options = {}) {
     if (typeof mjml === 'string') {
-      if (mjml.indexOf('<mjml>') === -1) {
-        // supports returning siblings elements from a custom component
-        const partialMjml = MJMLParser(`<mjml><mj-body>${mjml}</mj-body></mjml>`, {
-          ...options,
-          components,
-          ignoreIncludes: true,
-        })
-        const body = find(partialMjml.children, { tagName: 'mj-body' })
-
-        return body.children
-          .map(child => this.context.processing(child, this.context))
-          .join('')
-      }
-
-      mjml = MJMLParser(mjml, {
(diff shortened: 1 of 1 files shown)


kmcb777 (approved):

packages/mjml-core/src/createComponent.js
```diff
@@ -72,6 +72,20 @@ class Component {
 
   renderMJML(mjml, options = {}) {
     if (typeof mjml === 'string') {
+      if (mjml.indexOf('<mjml>') === -1) {
```
iRyusa, line 75: I'm not sure that we need that if here ? 

We can just wrap it into a "safe" tag like `<fragment>` (to keep React-ish vocabulary) and just map over children as you do l 84.

It should be way simpler and prevent later `find`

IanEdington: I'll give that a try

kmcb777: i'm not sure if we have any "safe" tag like this that is parseable and doesn't add anything
but you could juste wrap in `mj-body`, then you don't need the `find` because your `partialMjml` will already be the body

IanEdington:
# Testing from conversation in Slack

> Since it will only affect custom components, it shouldn't break any existing templates, but it will affect ALL custom components since probably none of them renders something containing <mjml>. It would be worth testing if functionalities like componentHeadStyle still works. Also, if a custom component is inside a section and renders an array of columns, it would be nice to check if the columns width calculation still works properly.
> 
> We'll also need to test if a custom component that uses mj-head components still work fine. I think it should be ok but it's worth checking. And we'll need to test all of them since most of them work in different ways

packages/mjml-core/src/createComponent.js
iRyusa: I think it's just to ensure that we use a known vocabulary `fragment` is a known & used term instead of mj-body which would make you think that you're starting a new document 🤔

iRyusa:
@IanEdington Do you have some time to look at this so we can merge this for next release ?

IanEdington:
I took a stab at testing this but the work started to get quite big.
It seems like if testing is something we want to do more of it might be worth investing in some tooling improvements, which seem like they are outside the scope of this PR.
How important is it to write automated tests for this change?

I unfortunately won't have time this week to do automated tests for this change but would be able to implement the suggestions and do manual testing on a couple cases.

Also, feel free to take over this PR

IanEdington:
Sounds good

packages/mjml-core/src/createComponent.js
IanEdington: It's working with `fragment`. I like that it's not an existing `mj-` component so it's clear that we're only wrapping for the children.

IanEdington:
## I've tested with:
- A component returning a single node
- A component returning multiple nodes
- A component containing an mj-head

## I did not test:
- a custom component is inside a section and renders an array of columns (columns width calculation)
I don't have enough context on how it works to know if it's working properly

IanEdington:
@iRyusa @kmcb777 Just wanted to ping you on this in case there's more that is needed

iRyusa:
We're just a bit busy, we'll get this merged in 4.8 👍

iRyusa (approved):

IanEdington:
🙌 No worries 😄

PR #2049, opened 2020-10-02
Fix issue with stacking columns within mj-group in Gmail for IOS 

jsteele883:
This PR resolves the bug in Gmail on IOS where the `word-spacing: 1px` is added to the body causing the mj-group columns to stack. The issue is outlined here https://github.com/hteumeuleu/email-bugs/issues/80 and the suggested fix is provided by Mark Robbins.

Example MJML to reproduce issue:

```
<mjml>
  <mj-body>
    <mj-section>
      <mj-group>
        <mj-column>
          <mj-text font-size="13px" color="#000000">Text left</mj-text>
        </mj-column>
        <mj-column>
          <mj-text font-size="13px" color="#000000">Text right</mj-text>
        </mj-column>
      </mj-group>
    </mj-section>
  </mj-body>
</mjml>
```

Columns are stacking in IOS 14 on Gmail. The fix will ensure that the columns remain inline. 

Email on Acid link testing updated MJML https://app.emailonacid.com/app/acidtest/dl3EjaVeiHMeg4jQuqi7vci31erb7xGdGJa93CSOY0QZ6/list

### DIFF
--- a/packages/mjml-core/src/helpers/skeleton.js
+++ b/packages/mjml-core/src/helpers/skeleton.js
@@ -76,10 +76,9 @@ export default function skeleton(options) {
         }
         ${headRaw.filter(negate(isNil)).join('\n')}
       </head>
-      <body${
-        backgroundColor === ''
-          ? ''
-          : ` style="background-color:${backgroundColor};"`
+      <body${backgroundColor === ''
+        ? ' style="word-spacing:normal"'
+        : ` style="word-spacing:normal;background-color:${backgroundColor};"`
       }>
         ${buildPreview(preview)}
         ${content}
--- a/packages/mjml-core/src/helpers/skeleton.js
+++ b/packages/mjml-core/src/helpers/skeleton.js
@@ -76,10 +76,7 @@ export default function skeleton(options) {
         }
         ${headRaw.filter(negate(isNil)).join('\n')}
       </head>
-      <body${backgroundColor === ''
-        ? ' style="word-spacing:normal"'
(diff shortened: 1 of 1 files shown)


packages/mjml-core/src/helpers/skeleton.js
```diff
@@ -76,10 +76,9 @@ export default function skeleton(options) {
         }
         ${headRaw.filter(negate(isNil)).join('\n')}
       </head>
-      <body${
-        backgroundColor === ''
-          ? ''
-          : ` style="background-color:${backgroundColor};"`
+      <body${backgroundColor === ''
+        ? ' style="word-spacing:normal"'
```
iRyusa, line 80:
```
style="word-spacing:normal;${backgroundColor ? `background-color:${backgroundColor};` : ''}"
```

Maybe something like that to avoid repetition ?

iRyusa:
That's so good 👍 Thanks for this

packages/mjml-core/src/helpers/skeleton.js
jsteele883: @iRyusa Thanks for the suggestion I've updated that now

iRyusa (approved):

iRyusa:
Maybe we can remove the note about minify on mj-group then ?

jsteele883:
> Maybe we can remove the note about minify on mj-group then ?

It could be the same bug. I don't have an IOS9 device to test so I couldn't say for certain. Let me know what you think and I can remove the note if necessary.

PR #2051, opened 2020-10-06
mjml-hero add border-radius

vasin:


iRyusa:
Hello, thanks for your PR have you done some test on compatibility for both modes ?

vasin:
> Hello, thanks for your PR have you done some test on compatibility for both modes ?

Checked the display and in the mode "fluid-height/fixed-height"

PR #2059, opened 2020-10-09
Move js-beautify and html-minifier to cli

TrySound:
Ref https://github.com/mjmlio/mjml/issues/1947

- minify outlook conditionals unconditionally like merging
- moved js-beautify and html-minifier to cli
- deprecated "minify" and "beautify" options in mjml-core

### DIFF
--- a/packages/mjml-cli/package.json
+++ b/packages/mjml-cli/package.json
@@ -28,6 +28,8 @@
     "@babel/runtime": "^7.8.7",
     "chokidar": "^3.0.0",
     "glob": "^7.1.1",
+    "html-minifier": "^3.5.3",
+    "js-beautify": "^1.6.14",
     "lodash": "^4.17.15",
     "mjml-core": "4.7.1",
     "mjml-migrate": "4.7.1",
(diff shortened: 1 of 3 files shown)


TrySound:
сс @kmcb777 @iRyusa @DRoet

iRyusa:
Looks good to me cc @ngarnier 5.X branch will remove those from core maybe we should discuss if we have more breaking changes to push in it  ?

ngarnier:
Ok yeah @iRyusa let's chat this week

iRyusa (approved):

TrySound:
I think we can provide custom plugins with hooks which run any minification or other tool. I can implement it in separate PR and default to it in cli.

iRyusa:
That would be so cool 👍 I'm merging this one as it's fine for MJML 4.8 then

TrySound:
Will do this in a couple of weeks after vacation.

IanEdington:
@TrySound It seems like this will this break minification in the gulp tool. Is that true?

iRyusa:
Nope it wont the option will still be handled in core but display a warning during the transition. We'll just add those into gulp-mjml dependencies that's not that hard

TrySound:
Or piped gulp plugins can be used instead to not bloat gulp-mjml and make it single purpose.

IanEdington:
@iRyusa Thanks for pointing that out 👍 

@TrySound I like the idea of simplifying MJML by removing these dependencies. That approach seems like a good one for a v5 change.

PR #2060, opened 2020-10-09
Add mjml-preset-core package

TrySound:
Ref https://github.com/mjmlio/mjml/issues/1947

This is similar take to https://github.com/mjmlio/mjml/pull/1950 but
preset is distributed as an object with components array and
dependencies tree instead of register function. This allows us to
pass this preset to mjml2html and maybe eventually get rid from globally
mutated state.


TrySound:
сс @kmcb777 @iRyusa @DRoet

iRyusa:
I'm a bit concerned about the perf implication to register every component on each mjml2html invoke here ? Shouldn't be really big of course... but i'm not sure that the most efficient way to do this

TrySound:
You are right. I added deduplication logic to registerDependencies. registerComponent does not allow duplication.
The point of this change is getting rid from global configuration so we could allow running mjml with different presets. This can be useful with custom components.

iRyusa (approved):
Looks good to me, as it breaks custom component registration we should wait to release this with a MJML5 tho

iRyusa:
...I'm so bad at reading stuff `registerComponent` is still called. Sorry then we should be fine if we can resolve those conflicts

🙇

TrySound:
Could yiu please do this? Im away from computer for the next two weeks. Mobile github is not very good.

iRyusa:
👍 I'll get this done thanks again for your awesome contributions

iRyusa:
Oops, merging the validator revamp did some conflicts here. 

We might want to add a small section about preset-usage in our doc tho ( https://github.com/mjmlio/mjml/blob/master/doc/config.json ) ?

TrySound:
I think it's ready

DRoet (approved):
LGTM, @iRyusa I think this one can be merged?

TrySound:
@kmcb777 Rebased

kmcb777:
ah perfect thanks

PR #2066, opened 2020-10-21
Mj-button-clickable; docs only

GarryFlemings:
This is a simple notice that mj-button won’t be fully clickable as there’s no bulletproof solution. It includes a link to Issue #359.

file: only packages/mjml-button/README.md
no JavaScript

### DIFF
--- a/packages/mjml-button/README.md
+++ b/packages/mjml-button/README.md
@@ -6,6 +6,14 @@
 
 Displays a customizable button.
 
+<aside class="notice">
+  The `mj-button` component is almost fully clickable.
+  If we improve capability for one supported email client, we'd harm support for another.
+  <br /><br />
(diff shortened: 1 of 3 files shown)


packages/mjml-button/README.md
```diff
@@ -6,6 +6,11 @@
 
 Displays a customizable button.
 
+<aside class="notice">
+  The `mj-button` won't be fully clickable because of client support.
+  See discussion at [Issue #359](https://github.com/mjmlio/mjml/issues/359).
```
swrobel, line 11: This Markdown doesn't actually render a link on the [docs page](https://documentation.mjml.io/#mj-button): 
Screen Shot 2021-04-07 at 5 34 50 PM

PR #2070, opened 2020-11-02
document rel attribute for mj-social-element.

KonstantinLukaschenko:
This PR adds the missing documentation for the `rel` attribute on `mj-social-element`.


iRyusa:
👍 Thanks

PR #2071, opened 2020-11-02
Revamp validator

TrySound:
- drop meaningful warning package
- statically define builtin rules (avoided iterating module namespace)
- drop lodash because builtin methods are good enough
- pass dependencies to rule options instead of importing in rule module
- implement assignDependencies utility based on dependencies structure
- fixed duplicating dependencies when registering multiple times
- require passing dependencies explicitly to get rid from side effects (warn for now)

### DIFF
--- a/packages/mjml-validator/src/dependencies.js
+++ b/packages/mjml-validator/src/dependencies.js
@@ -1,6 +1,6 @@
 export const assignDependencies = (target, ...sources) => {
   for (const source of sources) {
-    for (const tag of source) {
+    for (const tag of Object.keys(source)) {
       const list = []
       if (target[tag]) {
         list.push(...target[tag])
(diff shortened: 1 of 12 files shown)


packages/mjml-validator/src/dependencies.js
```diff
@@ -1,15 +1,25 @@
...
+    }
+    if (right[tag]) {
+      list.push(...right[tag])
+    }
+    newDependencies[tag] = Array.from(new Set(list))
   }
+  return newDependencies
 }
 
 const dependencies = {}
```
iRyusa, line 19: As you reassign dependencies, it must be a `let` instead no ?

iRyusa (approved):
Looks good to me, can we fix the CI ?

TrySound:
Looks like babel is not able to transpile with live bindings so mutating dependencies is the only way for now.
I replaced mergeDependencies with assignDependencies which can be used as immutable with empty object as first parameter.

PR #2072, opened 2020-11-02
Transpile lib for node 10

TrySound:
This change allows to not add helpers and regenerator for many features
which work in node and modern web.

Disabled syntax restriction in eslint to allow using for/of loops.


iRyusa:
👍

Issue #407, opened 2016-10-07
Suggestion: configurable alignment for divider


zauni:
Hey MJML Team,
first of all: thanks for this awesome framework :) Really great job!

And now the problem: I want to have a 80px wide divider left aligned, which is not possible today. Would it be a good idea to let the margin attribute be configurable? like this

```
<mj-divider margin="0" width="80" />
```

Or could you guide me to the right direction on how to extend the mj-divider component with my own component which has this attribute? I already tried to extend the Divider class and overwrite the getStyles method, but didn't work out as expected. Now I just copied the Divider class and put the margin attribute on it, but this is not so nice solution.

And if I want to make a pull request for the original divider class, what's the best way to support "margin" and "margin-(left|right|top|bottom" attributes at the same time? Is there a utility method for this kind of stuff?

Thanks for your help!

ngarnier:
Hey @zauni, 

Thanks for the feedback and for reporting this issue! Here are 2 workarounds to achieve what you want:

1. Add some padding to force the divider to go left, using `padding-right`
2. Use columns and insert the divider in the left column of your section

Regarding pull-requests to add support for new attributes, here is an example of a similar contribution that could help you: https://github.com/mjmlio/mjml/pull/298 (but here, I don't think we should add `margin` as there is `padding` already).

zauni:
Hi @ngarnier,
thanks for the reply and your 2 suggestions. The first workaround is unfortunately not a good solution on smaller screens (< 600px) because a horizontal scrollbar appears.

example:
```html
<mj-divider border-color="blue" border-width="3px" width="80px" padding-right="495px" />
```

with the following code it is working okay on desktop, but not perfect on mobile (just 32px wide on 320px screen width):
```html
<mj-divider border-color="blue" border-width="3px" width="100%" padding-right="82%" />
```

The column workaround has almost the same problems with the percentage widths, but it is 100% wide on smaller screens.

So maybe an option to switch off the auto margins wouldn't be so bad? :)

ngarnier:
Regarding the columns technique, you have to use `mj-group` (https://mjml.io/documentation/#mjml-group) to prevent the columns from stacking on mobile. 

We'll explore the auto margins, in the meantime let met know if it helps!

zauni:
@ngarnier okay, with the group tag it's better, but also too narrow an small screens:

```html
<mj-section background-color="#fff">
    <mj-group>
        <mj-column width="22%">
[...]

PR #2087, opened 2020-11-11
Add align attribute to divider

thierrylemoulec:
Hi! I'd like to add align attribute for `<mj-divider />`, so I created this pull request.

I can't include litmus/email on acid screenshot because I don't have an account. But if there is a compatibility issue I'd be happy to work on it and provide a working version for this simple feature.

This PR would help resolve this issue #407 without using a workaround.

### DIFF
--- a/packages/mjml-divider/src/index.js
+++ b/packages/mjml-divider/src/index.js
@@ -29,17 +29,18 @@ export default class MjDivider extends BodyComponent {
   }
 
   getStyles() {
+    let computeAlign = '0px auto'
+    if (this.getAttribute('align') === 'left') {
+      computeAlign = '0px'
+    } else if (this.getAttribute('align') === 'right') {
+      computeAlign = '0px 0px 0px auto'
+    }
     const p = {
       'border-top': ['style', 'width', 'color']
         .map((attr) => this.getAttribute(`border-${attr}`))
         .join(' '),
       'font-size': '1px',
-      margin:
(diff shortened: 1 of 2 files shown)


iRyusa:
This need to be tested first cc @kmcb777

packages/mjml-divider/src/index.js
```diff
@@ -16,6 +16,7 @@ export default class MjDivider extends BodyComponent {
     'padding-right': 'unit(px,%)',
     'padding-top': 'unit(px,%)',
     width: 'unit(px,%)',
+    align: 'enum(left, center, right)',
```
kmcb777, line 19: Here there must be no spaces before 'center' and 'right', otherwise the validator expects those spaces

kmcb777 (changes requested):

kmcb777:
Hi @thierrylemoulec thanks for this PR !
Could you remove the spaces in the allowedAttributes, cf. my previous comment
One more thing is needed for this to work in outlook :
In `renderAfter` on line 84, `align: 'center'` needs to be replaced with `align: this.getAttribute('align'),`
With this fix it seems to work perfectly everywhere 👍

kmcb777:
i'd like to include this in tomorrow's release so i'll merge this and add the little fix

thierrylemoulec:
@kmcb777 Sorry for the delayed response. I wasn't available. Thanks for doing the fix and merging the feature. 👍

PR #2100, opened 2020-11-27
Update index.js

Pharmasolin:
Added missing <tbody> after <table> tag in mj-column component.

----
It's better to fix issue with missing <tbody> after <table> tag now.

As with new feature `<mj-html-attributes>` users might try to target tags inside mj-column with child combinator `>` and as there is no <tbody> inside they might see an issue where in Google Chrome inspector they will see <tbody> tag in HTML, but in actual html file this tag would be missing.

@iRyusa you might remember conversation regarding this situation in slack.


iRyusa:
👍 ty for that

PR #2101, opened 2020-11-30
Mj-section_cut_extra_aside

GarryFlemings:
A "warning" aside is duplicated early in **packages/mjml-section/README.md**.

This cuts the extra one.

Further, it shortens the wording in that aside. Also, it removes awkward wording and clarifies.

We better communicate the same message.

No change to any JavaScript file.


GarryFlemings:
Flawed.

PR #2116, opened 2020-12-18
Browser build

kmcb777:


packages/mjml-browser/browser-mocks/path.js
```diff
@@ -0,0 +1,12 @@
+const mockFn = () => {
```
iRyusa, line 1: Can we add eslint ignore here ?

DRoet:
Awesome!

kmcb777:
This will be included in the next release (4.8) in its current state, then when v5 is released without minify/beautify, the build will get smaller

PR #2122, opened 2020-12-25
Fixed background image bug in Yahoo and AOL (mjml-hero) closes #2098

### LINKED ISSUE #2098: Background-image for Yahoo mail
When compiling, the background is set without quotes. Such background rejects Yahoo mail. How to get single quotes during compilation. Or how to display a background for Yahoo.
Thanks!
Screenshot_1

should be:

```
style="background: url('https//.....')..."
```

Pharmasolin:
AOL and Yahoo doesn't show background image if url() function doesn't receive an image in quotes.

Test result:
https://litmus.com/pub/c96d831

Also I noticed that `getBackground()` returns shorthand `background` style, and then `getStyles()` in `hero` object  returned again `background-position` and `background-repeat` which I removed not sure if that was for some hack or no, but test shows that all is ok.

### DIFF
--- a/packages/mjml-hero/src/index.js
+++ b/packages/mjml-hero/src/index.js
@@ -99,8 +99,6 @@ export default class MjHero extends BodyComponent {
       },
       hero: {
         background: this.getBackground(),
-        'background-position': this.getAttribute('background-position'),
-        'background-repeat': 'no-repeat',
         'border-radius': this.getAttribute('border-radius'),
         padding: this.getAttribute('padding'),
         'padding-top': this.getAttribute('padding-top'),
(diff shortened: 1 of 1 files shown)


kmcb777:
After a few tests on emailonacid, background image seemed to work already without your fix, but the content of the hero wasn't displayed on AOL and Yahoo, and now it is, so let's merge this anyway 👍 
Also i didn't witness any regression on background-position and background-repeat

kmcb777:
actually after a few more tests, those attributes seem useful, mainly for Aol and Yahoo. They are part of the few clients that seem to totally ignore vertical background-position, but if we completely remove it here then on a mj-hero that's higher than the image, the background will be at the bottom. And in this case the background-repeat is useful as well on Aol and Yahoo.
I'd rather not change this behavior for existing templates, so i'll keep the added quotes but i'll re-add the attributes in another PR

Issue #1891, opened 2020-05-01
Support borderTop, borderBottom, etc... on mjml-image


osdiab:
**Is your feature request related to a problem? Please describe.**
You can set the `border` attribute on `mjml-image`s but you can't set specific ones like `border-top` or `border-bottom`. 

**Describe the solution you'd like**
Add those props to `mjml-image` and probably other elements that also have border support

**Describe alternatives you've considered**
In my case using `mjml-style` is not convenient since I'm making templates that can be used on multiple emails.

**Additional context**
I am using `mjml-react` with TypeScript, so if this gets implemented, that library, plus typings for `mjml` and `mjml-react`, will need to be updated too

iRyusa:
We're not maintainer of typescript definition so you'll have to check with them for that (probably using the allowedattribute so it should be fine)

cc @kmcb777 we should be able to add it for next release

osdiab:
@iRyusa i can add it there, no worries. will do whenever this gets resolved!

iRyusa:
👍 Done and released in 4.8

PR #2124, opened 2020-12-25
Fixed independent border properties, closes #1939

### LINKED ISSUE #1939: MjImage: border-left/top/right/bottom does not work
In `getStyles()` there is `'border-left': this.getAttribute('left')`, but the `left` attribute is not declared in `allowedAttributes`. According to @iRyusa this should be `'border-left': this.getAttribute('border-left')`.

https://github.com/mjmlio/mjml/blob/master/packages/mjml-image/src/index.js#L59

Pharmasolin:
Documentation on https://mjml.io/documentation/#mj-image might needed to be updated.

### DIFF
--- a/packages/mjml-image/src/index.js
+++ b/packages/mjml-image/src/index.js
@@ -56,10 +56,10 @@ export default class MjImage extends BodyComponent {
     return {
       img: {
         border: this.getAttribute('border'),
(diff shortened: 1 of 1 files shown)


Pharmasolin:
Also related to #1891

PR #2135, opened 2021-01-06
Made calculation changes within index.js

quetzalcoatlAxolotl:
Noticed divider width was being parsed incorrectly within outlook conditional tables (ie: `width:6%` was being parsed as `width:-64px` for outlook).

Corrected calculation for `getOutlookWidth` within index.js so email padding is subtracted from the email width rather than the parsed width.

### DIFF
--- a/packages/mjml-divider/src/index.js
+++ b/packages/mjml-divider/src/index.js
@@ -67,9 +67,9 @@ export default class MjDivider extends BodyComponent {
 
     switch (unit) {
       case '%':
-        const effectiveWidth = parseInt(containerWidth, 10) - paddingSize;
-        const percentMultiplier = parseInt(parsedWidth, 10) / 100;
-        return `${effectiveWidth * percentMultiplier}px`;
+        const effectiveWidth = parseInt(containerWidth, 10) - paddingSize
+        const percentMultiplier = parseInt(parsedWidth, 10) / 100
+        return `${effectiveWidth * percentMultiplier}px`
       case 'px':
         return width
       default:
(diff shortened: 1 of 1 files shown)


iRyusa:
@kmcb777 can you review this one too cf https://mjml.io/try-it-live/tlznqVxR3b3

kmcb777:
The calculation seems right and the result is good 👍 

@quetzalcoatlAxolotl thanks for this !
Eslint complains about the semicolons and declarations not being in a block, so could you just replace with this plz : 
```
      case '%': {
        const effectiveWidth = parseInt(containerWidth, 10) - paddingSize
        const percentMultiplier = parseInt(parsedWidth, 10) / 100
        return `${effectiveWidth * percentMultiplier}px`
      }
```

quetzalcoatlAxolotl:
> The calculation seems right and the result is good 👍
> 
> @quetzalcoatlAxolotl thanks for this !
> Eslint complains about the semicolons and declarations not being in a block, so could you just replace with this plz :
> 
> ```
>       case '%': {
>         const effectiveWidth = parseInt(containerWidth, 10) - paddingSize
>         const percentMultiplier = parseInt(parsedWidth, 10) / 100
>         return `${effectiveWidth * percentMultiplier}px`
>       }
> ```

Thank you for the feedback! I'm new to committing to mjml, so I was curious about that. I'll make those changes now :)

kmcb777:
@quetzalcoatlAxolotl thanks again, however you removed the semicolons but not added the `{ }` around the case block

quetzalcoatlAxolotl:
> @quetzalcoatlAxolotl thanks again, however you removed the semicolons but not added the `{ }` around the case block

Raaah hahah thanks, third time's a charm.

kmcb777:
Thanks, perfect :)

PR #2150, opened 2021-01-22
Reduce global components usage

TrySound:
In https://github.com/mjmlio/mjml/pull/2060 I want to pass `presets` without mutating global components.
In this diff I changed components map to be consumed from context in
component implementations.


iRyusa (approved):

iRyusa:
I'm just wondering if it will not increase memory consumption as the `components` object will be copied over and over?

TrySound:
In this particular PR only reference to object is passed. In preset PR not a big deal. It's not an object with millions or thousands fields. Multiple cheerio invocations consume much more memory.

PR #2153, opened 2021-01-22
Default dependencies in mjmlconfig

### LINKED ISSUE #2141: TypeError: Cannot convert undefined or null to object
**Description**
When registering a custom component that does not contain a `dependencies` property, the `mjml-validator` gives a very cryptic error,
```
Error when registering custom component :  /path/to/CustomComp.js TypeError: Cannot convert undefined or null to object
```

**To Reproduce**
1. Create a custom mjml component as the default export
```js
export default CustomComp extends BaseComponent {
  // public dependencies = {}; Leave this property out
  public allowedAttributes = {
    attr1: 'string'
  };

  // ...
}
```
2. Add the component to your `.mjmlconfig`
3. Create an mjml template that uses your custom component
4. Compile the mjml with the node SDK
```js
import mjml2html from 'mjml';

const { html, errors } = mjml2html(mjmlFile);

console.log(errors);
```

**Expected behavior**
One of the following...

1. Allow a component to be defined and validated without specifying dependencies.
2. Throw a more descriptive error when registering dependencies that don't exist.

 - OS: MacOS
 - MJML Version: 4.8.1
 - MJML tool used: Node

**Additional context**
[This line of code](https://github.com/mjmlio/mjml/blob/master/packages/mjml-core/src/helpers/mjmlconfig.js#L95) is where you should handle either 
1. passing an empty object if `.dependencies` is not defined
2. throwing an error telling you that `dependencies` are required.

Without this, the `sources` param of `assignDependencies` ends up being `[null]`, which throws a hard type error when trying to iterate over the items.
https://github.com/mjmlio/mjml/blob/master/packages/mjml-validator/src/dependencies.js#L1

ryanburr:
Fixes #2141 

- Defaults the dependencies of a component to an empty object
- Adds type checks and warnings to dependency assignment

### DIFF
--- a/packages/mjml-validator/src/dependencies.js
+++ b/packages/mjml-validator/src/dependencies.js
@@ -1,5 +1,5 @@
 export const assignDependencies = (target, ...sources) => {
-  if (!sources || !sources.length) {
+  if (sources.length === 0) {
     return target
   }
 
--- a/packages/mjml-validator/src/dependencies.js
+++ b/packages/mjml-validator/src/dependencies.js
@@ -1,14 +1,26 @@
 export const assignDependencies = (target, ...sources) => {
+  if (!sources) {
+    return target
+  }
+
   for (const source of sources) {
-    for (const tag of Object.keys(source)) {
-      const list = []
-      if (target[tag]) {
-        list.push(...target[tag])
+    if (typeof source === 'object' && source !== null) {
+      for (const tag of Object.keys(source)) {
+      if (typeof tag !== 'string') {
+          const list = []
+          if (target[tag]) {
+            list.push(...target[tag])
+          }
+          if (source[tag]) {
+            list.push(...source[tag])
+          }
+          target[tag] = Array.from(new Set(list))
+        } else {
+          console.warn('dependency "tag" must be of type string')
+        }
       }
-      if (source[tag]) {
-        list.push(...source[tag])
-      }
-      target[tag] = Array.from(new Set(list))
+    } else {
+      console.warn('"dependencies" must be an object.')
     }
   }
   return target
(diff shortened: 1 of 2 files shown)


packages/mjml-validator/src/dependencies.js
```diff
@@ -1,14 +1,26 @@
...
-    for (const tag of Object.keys(source)) {
-      const list = []
-      if (target[tag]) {
-        list.push(...target[tag])
-      }
-      if (source[tag]) {
-        list.push(...source[tag])
+    if (typeof source === 'object' && source !== null) {
+      for (const tag of Object.keys(source)) {
+      if (typeof tag !== 'string') {
```
TrySound, line 9: Inverse check and fix formatting please

ryanburr: Good catch.

iRyusa:
Looks good to me @TrySound are you OK too ?

packages/mjml-validator/src/dependencies.js
```diff
@@ -1,14 +1,26 @@
 export const assignDependencies = (target, ...sources) => {
+  if (!sources) {
```
TrySound, line 2: Better check length. `sources` is always an array

TrySound:
Last small tweak

packages/mjml-validator/src/dependencies.js
ryanburr: Done.

packages/mjml-validator/src/dependencies.js
```diff
@@ -1,14 +1,26 @@
 export const assignDependencies = (target, ...sources) => {
+  if (!sources || !sources.length) {
```
TrySound, line 2: I mean `if (sources.length === 0)`

ryanburr: Done

PR #2176, opened 2021-02-11
Add support for title attribute in mj-button

### LINKED ISSUE #2175: mj-button is missing title
There is no title attribute for mj-button. This would be helpful to have in order to show a long tooltip for users as well as for accessibility.

Is there a technical reason why this was not implemented?

lahdekorpi:
This is added to the last element, a or p as the title attribute similarly to name. Closes #2175


lahdekorpi:
Not sure if a version bump or other stuff related to maintaining is required in this project.

iRyusa:
We'll bump everything on release thanks for this !

Issue #2157, opened 2021-01-27
"mj-accordion-element" supports "border" strangely


GarryFlemings:
**Describe the bug**

- Whether using attribute `border` directly on an `mj-accordion-element` or through `mj-attribute` (as in the example mentioned below), borders change consistent with each `mj-accordion-text` overriding its `border-top` and `border-bottom`, thus changing all horizontal lines for the `mj-accordion` component except the top line.
- `mj-accordion-element` (https://documentation.mjml.io/#mj-accordion) does not list `border` among its supported attributes.
- https://mjml.io/try-it-live/gs3v2xeDrGl includes the `border` attribute on `mj-accordion-element`. I used it accidentally and missed the notice, `attribute border is illegal`.

**To Reproduce**
1. https://mjml.io/try-it-live/gs3v2xeDrGl (version 4.7.1)
2. MJML app (version ^4.8.1)
3. Node v12.18.4 and MJML version 4.8.1.
![image](https://user-images.githubusercontent.com/14095071/106055774-77305200-60b3-11eb-8d90-6d059b1922c2.png)

**Expected behavior**
MJML would be better if either

![image](https://user-images.githubusercontent.com/14095071/106056314-25d49280-60b4-11eb-8b86-42a3ee2b962d.png)

1. The code would ignore unsupported attributes (`border` on `mj-accordion-element`, in this case), or
2. We change the documentation to indicate support for `border` on `mj-accordion-element`.

Recommended: Option 1 above and not Option 2. Ignore unsupported attributes and leave the documentation as is. It's probably a rare author that wants the existing rendering. I don't have a good proposal for this attribute here.

**MJML environment (please complete the following information):**
current Windows 10; MJML versions above

**Additional context**
If the decision is to change the documentation, I volunteer to do it.

iRyusa:
You can just add it to the doc, as it's seems intentional from the code.

GarryFlemings:
Didn't expect that, but I'll put in the PR.

Maybe you'll want to disarm the syntax warning "Attribute border is illegal" for `accordion-element` before merging that PR?

iRyusa:
You're right you can also add it to supported attribute

PR #2180, opened 2021-02-15
Accordion-docs-3

GarryFlemings:
Changes for mjml-accordion, both documentation and JavaScript.

### This PR

Adds attribute `border` to the list of attributes `accordion-element` supports, both in documentation (README.md) and in JavaScript (AccordionElement.js; `allowedAttributes`). Related discussion #2157 and #2159.

Also, though previous authors have done a nice job creating this documentation, there are tweaks here to make the documentation still more useful. I explain my reasons for each because it seems my suggestions are better accepted when I express those reasons.

Improves some grammar.
Improves some awkward wording and shortens some sentences.

[accordion-docs-3.diff.txt](https://github.com/mjmlio/mjml/files/5983306/accordion-docs-3.diff.txt)
(remove .txt extension for .diff file)

### border and mj-accordion-element Issue #2157
MJML documentation for mj-accordion-element does not indicate support for attribute border. However, using border there changes the rendering of all horizontal borders in the accordion except the top border. (An unusual effect.)

The structure of the code is assessed as indicating this status is intentional, so we're changing the documentation. In the documentation, see the attribute list for mj-accordion-element > border.

### Grammar
An `aside` in mj-accordion uses sentence structure, "Every attribute in mj-accordion are applied". The verb could be better.

The text in mj-accordion-element mentions "a accordion pane". The article could be better.

### Expression
The aside in mj-accordion (already mentioned) is about attributes in mj-accordion and mj-accordion-element, as well as about effects in mj-accordion-element. A more natural placement is with mj-accordion-element. The proposed location is with mj-accordion-element.

A sentence structure used in mj-accordion-element, mj-accordion-title, and mj-accordion-text includes, "This component enables you". Both pairs of words lengthen the sentences without adding content. See the shorter phrases in each section.

The text in mj-accordion-title is a compound sentence. It's awkward because when broken into two sentences, one is, "This component enables you to style a title to your accordion." For this half of the sentence, "in" is better than "to", but not for the other half. The sentence could be better. See new text there.

The text in mj-accordion-element introduces a term--"accordion pane"--that is not defined and never used again. That can be better. I propose using "title/text pair" instead. See related changes in mj-accordion-element, mj-accordion-title, and mj-accordion-text.

### Slight addition

In current documentation, it's unsaid why mj-accordion-element exists. It's reasonably clear why the others exist. If we can find a short sentence to introduce the concept, the documentation improves. I propose one.

### DIFF
--- a/packages/mjml-accordion/README.md
+++ b/packages/mjml-accordion/README.md
@@ -48,16 +48,11 @@
   <a href="https://mjml.io/try-it-live/components/accordion">
     <img width="100px" src="https://mjml.io/assets/img/svg/TRYITLIVE.svg" alt="sexy" />
   </a>
-</p>
-
-<aside class="notice">
-Every attribute in `mj-accordion` are applied to `mj-accordion-element` unless you override them on `mj-accordion-element`
-</aside>
 
 
 attribute | unit | description | default value
 ----------|------|-------------|---------------
-border | n/a | border | n/a
+border | string | CSS border format | none
 container-background-color | n/a | background-color of the cell | n/a,
 css-class | string | class name, added to the root HTML element created | n/a
 font-family | n/a | font | Ubuntu, Helvetica, Arial, sans-serif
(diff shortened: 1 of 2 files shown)


iRyusa:
@kmcb777 This should be ok to merge I think can you squash those commits to remove the merge commit ?

iRyusa (approved):

PR #2204, opened 2021-03-09
Remove beautify minify and migrate from mjml-core

kmcb777:


iRyusa:
We might need todo a 4.9.x branch as we've done with MJML 3, 2 and 1 before merging this

Issue #1581, opened 2019-04-18
Mailchimp is clearing/removing styles in head


mio31337:
My built HTML email templates were just another ones that I did for my client. Everything was smooth and perfect with the emails in all email clients (dekstop and mobile), until the client decided to switch to Mailchimp. The main issues was on Gmail App (iOS and Android). Everywhere else worked just fine as expected.

After a couple of hours scratching my head around, I finally found a bug that was causing Mailchimp to "strip" and in my opinion "comment out" all my styles and media queries from head. This caused the problem only for Gmail App.

I basically went on test-grinding and tested part by part of the source code from head until I finally came to a little piece of code "<!-->" that shouldn't even be there.

After I removed this opening HTML comment (usually on 2 spots after mjml generated my HTML file), I finally got it working in Mailchimp as expected. My styles in head were still there, mobile responsive layout was 100% there!

The code that MJML automatically adds and it should be removed manually:

![image](https://user-images.githubusercontent.com/722467/56363815-4c7e4d80-61ed-11e9-8cf3-9d5084e11233.png)

![image](https://user-images.githubusercontent.com/722467/56363855-6c157600-61ed-11e9-88af-9d81577e4dda.png)

Hopefully you get this fixed quickly.

iRyusa:
Well they shouldn't alter the HTML at all to be fair.

We use this solution as stated in this litmus topic, https://litmus.com/community/discussions/982-outlook-overrides-font-to-times-new-roman , so I believe this extra `<!-->` is needed ?

I don't know,  can you test a bit with litmus/eoa if it doesn't transform every text to Times when removing it ?

mio31337:
> Well they shouldn't alter the HTML at all to be fair.
> 
> We use this solution as stated in this litmus topic, https://litmus.com/community/discussions/982-outlook-overrides-font-to-times-new-roman , so I believe this extra `<!-->` is needed ?
> 
> I don't know, can you test a bit with litmus/eoa if it doesn't transform every text to Times when removing it ?

Yeah, I know. They shouldnt...

Since my post, I've been testing on all kinds of service providers - all looks perfectly fine without `<!-->`

Skjolbir:
> > Well they shouldn't alter the HTML at all to be fair.
> > We use this solution as stated in this litmus topic, https://litmus.com/community/discussions/982-outlook-overrides-font-to-times-new-roman , so I believe this extra `<!-->` is needed ?
[...]

PR #2221, opened 2021-03-31
fix invalid HTML comments in skeleton

FelixSchwarz:
This skeleton contains some HTML comments inside the conditional comment which I believe is illegal in HTML. This was added in 7fdd028d0e8b03e0bd73fddeec67d53f799be6fe but the additional comments seem to be an accidental change.

### DIFF
--- a/packages/mjml-core/src/helpers/skeleton.js
+++ b/packages/mjml-core/src/helpers/skeleton.js
@@ -29,9 +29,9 @@ export default function skeleton(options) {
         <title>
           ${title}
         </title>
-        <!--[if !mso]><!-->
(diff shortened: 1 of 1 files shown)


FelixSchwarz:
actually it seems somewhat intentional after reading https://en.wikipedia.org/wiki/Conditional_comment#Downlevel-revealed_conditional_comment  - still looks weird triggers the Python-based HTML parsers I'm using.

iRyusa:
It was intentional because of a post in email on acid I think @kmcb777 did some test with more "standard" MSO comment and it worked fine. Might close #1581 too

Issue #2215, opened 2021-03-19
Adding multiple font attributes causes email client to not import fonts


tonyketcham:
**Describe the bug**
This could potentially be a user error based on misunderstanding the docs. Using multiple <mj-font> tags in the head is causing just the default font to be used in the received email, client-side. The preview HTML shows up correctly though in iContact as well as the MJML editor: https://mjml.io/try-it-live/ACMGLQPDcho

**To Reproduce**
1. Create a JSON obj with multiple `<mj-font>` attributes in the `<mj-head>`. In markup that translates to:
```html
<mjml>
  <mj-head>
     <mj-font name="Raleway"
       href="https://fonts.googleapis.com/css?family=Raleway" />
     <mj-font name="Lobster"
       href="https://fonts.googleapis.com/css?family=Lobster" />
     <mj-font name="Balsamiq Sans"
       href="https://fonts.googleapis.com/css?family=Balsamiq%20Sans" />
  </mj-head>
  <mj-body>
    <mj-section>
      <mj-column>
        <mj-text font-size="20px" color="#F45E43" font-family="Raleway">Hello World</mj-text>
        <mj-text font-size="20px" color="#F45E43" font-family="Lobster">Hello World</mj-text>
        <mj-text font-size="20px" color="#F45E43" font-family="Balsamiq Sans">Hello World</mj-text>
      </mj-column>
    </mj-section>
  </mj-body>
</mjml>
```
2. Render it to HTML by doing `mjml2html` (I'm doing JSON in MJML but testing it in the editor with raw MJML gives the same issue)
3. Send the HTML to either Outlook or Gmail accounts

**Expected behavior**
All fonts should display in the email client

**MJML environment (please complete the following information):**
 - OS: MacOS
 - MJML Version: 4.9.0
 - MJML tool used: mjml2html (mjml-browser)

**Email sending environment(for rendering issues)**:
 - Platform used to send the email [iContact](https://icontact.com/)]  

**Affected email clients (for rendering issues):**
 - Email Client: Gmail, Outlook
 - OS: Windows, iOS, Mac
 - Browser: Chrome mobile/desktop, & in-app

**Screenshots**
Email Preview:
![image](https://user-images.githubusercontent.com/43280336/111824325-0f78d500-88b4-11eb-977b-92df1d807750.png)

Outlook web client received:
![image](https://user-images.githubusercontent.com/43280336/111824463-3fc07380-88b4-11eb-8a4c-b71242c52a89.png)

[...]

PR #2226, opened 2021-04-08
mj-font_external_fonts

GarryFlemings:
**Changes**

Only to packages/mjml-font/README.md

**Problem**

@tonyketcham, in Issue #2215 suggested additional information about client support of external fonts in "mj-font"

**Action**

Add an aside pointing to https://www.caniemail.com/features/css-at-font-face/

**Result**

Resolves Issue #2215


iRyusa:
Don't want to add a warning for each component. We probably need to think
of something else.

PR #2245, opened 2021-04-28
fix #2241 re-implement owa

### LINKED ISSUE #2241: Can the OWA attribute be implemented back?
I know about the [new version of OWA](https://github.com/mjmlio/mjml/issues/2097) but as long as OWA can be self-hosted it doesn't have to be up-to-date. All of my clients and my company uses older versions of OWA and I had to downgrade MJML lib to 4.8.2 in all of my projects. 

Is there any possibility to bring this attribute back as long as it is optional?

kmcb777:


doc/components_1.md
```diff
@@ -32,6 +32,7 @@ A MJML document starts with a `<mjml>` tag, it can contain only `mj-head` and `m
 
 attribute | unit | description | default value
 ----------|------|-------------|---------------
+owa | string | if set to "desktop", will force desktop display on outlook. Otherwise outlook will display mobile version since it ignores media queries | none
```
iRyusa, line 35: We should maybe update this "Switch displayed version for older (self-hosted) version of Outlook.com" And may be add a link to it #2241 ?

PR #2253, opened 2021-05-11
Fix background image in mj-section for Yahoo and AOL.

Pharmasolin:
Related to #2098 and #2122

As I understand it will also fix mj-wrapper with background images in Yahoo and AOL


iRyusa:
👍 thanks for that

PR #2263, opened 2021-05-22
If preview text exists don't display body text as preview

### LINKED ISSUE #1829: Create automatically white space after <mj-preview> text
Glad we have the <mj-preview> tag, but would be even better if it could by default apply "&nbsp;&zwnj;" after the text. More about this hack here:

https://litmus.com/blog/the-little-known-preview-text-hack-you-may-want-to-use-in-every-email

IanEdington:
closes #1829

We ran into an issue were preview text from the body of the email was being displayed after the preview text.
We found this article about how to prevent it and it's been working well for a couple months now.

This is based off this litmus article: https://www.litmus.com/blog/the-little-known-preview-text-hack-you-may-want-to-use-in-every-email/

Also thanks to @j56xie for her work finding this.

Example of the errors this fixes:
110679311-532d5980-81a5-11eb-9d2c-2b4911c240e0

### DIFF
--- a/packages/mjml-core/src/helpers/preview.js
+++ b/packages/mjml-core/src/helpers/preview.js
@@ -7,5 +7,8 @@ export default function (content) {
     <div style="display:none;font-size:1px;color:#ffffff;line-height:1px;max-height:0px;max-width:0px;opacity:0;overflow:hidden;">
       ${content}
     </div>
+    <div style="display:none;font-size:1px;color:#ffffff;line-height:1px;max-height:0px;max-width:0px;opacity:0;overflow:hidden;">
+      &nbsp;&zwnj;&nbsp;&zwnj;&nbsp;&zwnj;&nbsp;&zwnj;&nbsp;&zwnj;&nbsp;&zwnj;&nbsp;&zwnj;&nbsp;&zwnj;&nbsp;&zwnj;&nbsp;&zwnj;&nbsp;&zwnj;&nbsp;&zwnj;&nbsp;&zwnj;&nbsp;
+    </div>
   `
 }


packages/mjml-core/src/helpers/preview.js
```diff
@@ -7,5 +7,8 @@ export default function (content) {
     <div style="display:none;font-size:1px;color:#ffffff;line-height:1px;max-height:0px;max-width:0px;opacity:0;overflow:hidden;">
       ${content}
     </div>
+    <div style="display:none;font-size:1px;color:#ffffff;line-height:1px;max-height:0px;max-width:0px;opacity:0;overflow:hidden;">
```
IanEdington, line 10: This was copied from line 7. We've been using `display: none; max-height: 0px; overflow: hidden;` from the article so not sure about the differences between these two hide styles

iRyusa:
Will close #1829 

Thanks for this PR @IanEdington ! Can you gate this behavior behind an attribute instead of modifying the default behavior ?

IanEdington:
Cool thanks for the context. It seems like the kind of thing you would want on by default so new emails don't run into this weird behaviour. Would it make sense to include it as a breaking change in 4.9?

Are you thinking something like:

```
<Preview hideBody>
```

iRyusa:
I know that some local clients doesn't handle those chars properly and display them as plain text that's mostly why I see this as an "opt-in" feature.  Name looks good to me 👍

iRyusa:
They display directly the &zwnj; as a string in the preview text. It might
be related to the text-version as mentioned in goodemailcode. But I think
it behave like that on some french webmail too with no text-version sent.

renjiezhu2:
@IanEdington thank you for putting it up, we encountered the same issue as well. @iRyusa and is there a plan to merge this to master?

IanEdington:
There are a few issues with this PR. It only gives a small gap between the content. The more characters the more of a gap it provides. Because screens are getting wider it's theoretically not possible to put enough characters to always hide the other content.

From my perspective this needs someone to run a subject line test on litmus or something similar to understand how many characters are needed to provide a good enough gap.

I'm not going to get to this in the next few months. @renjiezhu2 feel free to take over this PR

IanEdington:
Hey all, I'm going to close this for now since I won't be working on it. I'd be happy for someone else to take over this work in another PR :D

renchris:
Is there any solution now to having the <mj-preview> text replace instead of prefix the body text in email preview text?

PR #2265, opened 2021-05-24
[DOC] Update mj-attributes docs

iRyusa:
- Add a ref to `mj-all`
- attributes -> MJML attributes
- Add a ref to `mj-class`


iRyusa:
cc @rootwork Does it seem less confusing ?

rootwork:
@iRyusa If I had to pick I would say #2266 is clearer, but this would be an improvement too.

iRyusa:
Let's take #2266 as a base and tweak it, closing this one then  👍
<https://emojipedia.org/thumbs-up/>

PR #2292, opened 2021-06-14
Feature request: Add a new attribute in order to get an outlook proof button

adrien-zinger:
## Purpose

In my current project I had to create a button with a border-radius.

I looked in the closed issue and I saw that you had never upgraded since this issue 2212. It might be some reason but I thought it was convegnant to create an attribute in the button that allow users to generate a vml button for outlook directly in mjml without mj-raw.

In this PR I just implemented another render function for the button (also fixed a mistaken align)
I added three attributes { mso-proof (boolean), mso-width, mso-height }.

The default behaviour is unchanged so it's backward compatible.

Test code:
```
<mjml>
  <mj-body>
    <mj-section>
      <mj-column>
        <mj-button font-family="verdana" border="2px solid #ccc000" mso-proof="true" border-radius="20px" background-color="#f45e43" color="white">
          Don't click me!
         </mj-button>
      </mj-column>
    </mj-section>
  </mj-body>
</mjml>
```

### DIFF
--- a/packages/mjml-button/README.md
+++ b/packages/mjml-button/README.md
@@ -13,6 +13,10 @@ Displays a customizable button.
     <a href="https://github.com/mjmlio/mjml/issues/359">Issue #359</a>.
 </aside>
 
+<aside class="warning">
+  The `mj-button` will be duplicated in two html components when mso-proof is true
+</aside>
+
 ```xml
 <mjml>
   <mj-body>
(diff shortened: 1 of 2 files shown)


iRyusa:
Hi !

Thanks for this PR but I don't think we'll want this to get merged in mjml-button for few reasons : 
- vml button can't be used inside a section with background image ( and it renders really bad in some cases )
- I believe this technique has some issue with multi line button too ?
- render consistency

However I don't want this work to be lost so what about create a repository and list it in the doc as community component ? 

You can put the md file here https://github.com/mjmlio/mjml/tree/de9df8a1a79be140e8f0c97db0447eb71c483067/doc and adds it to the config file here https://github.com/mjmlio/mjml/blob/de9df8a1a79be140e8f0c97db0447eb71c483067/doc/config.json

adrien-zinger:
Hi @iRyusa !

Nice, yes in fact the roundrect element is extremly rigid and there is a lot of deprecations. So I'll list it in the community component soon.

Thank you for your really fast answer !

Issue #2294, opened 2021-06-16
Role presentation missing on some tables


voermansm:
**Describe the bug**
For accessibility we are trying to add role="presentation" and found it was missing from in a few spots. I did review #2256 and that didn't seem to address this.

MJML Section
https://github.com/mjmlio/mjml/blob/d1879e81e16eff85f8a653094f9c066398bdee92/packages/mjml-section/src/index.js#L196

and 

MJML Table
https://github.com/mjmlio/mjml/blob/d1879e81e16eff85f8a653094f9c066398bdee92/packages/mjml-table/src/index.js#L78

(I hope that is a helpful way to indicate where/how to resolve missing attributes)

**To Reproduce**
https://mjml.io/try-it-live/mkjNKlEmR77
```
<mjml>
  <mj-body>
    <mj-section>
      <mj-column>

        <mj-image width="100px" src="/assets/img/logo-small.png"></mj-image>

        <mj-divider border-color="#F45E43"></mj-divider>

        <mj-text font-size="20px" color="#F45E43" font-family="helvetica">Hello World</mj-text>
				
        <mj-table>
          <tr>
            <td>No role on this table either</td>
          </tr>
        </mj-table>
      </mj-column>
    </mj-section>
       
  </mj-body>
</mjml>
```

**Expected behavior**
```
<!--[if mso | IE]>
      <table
         align="center" border="0" cellpadding="0" cellspacing="0" class="" style="width:600px;" width="600"
      >
```
becomes
```
<!--[if mso | IE]>
      <table
         align="center" role="presentation" border="0" cellpadding="0" cellspacing="0" class="" style="width:600px;" width="600"
      >
```

**MJML environment (please complete the following information):**
MacOS (but also happening on "Try it live"
 - MJML Version: 4.9.5 locally, 4.7.1 on Try it live
 - MJML tool used: CLI

**Affected email clients (for rendering issues):**
 - ADA / Screenreaders

iRyusa:
Hi !

I wasn't aware that it was needed on ghost table too ! Should be really easy to add 😄 

For mj-table we should add it as an mj-attribute with no default value because most of mj-table are use for proper tables !

voermansm:
I didn't realize it either. Then after an accessibility audit and realizing a screenreader isn't a device that consumes email, rather a proxy to whatever your email client is -- it made sense that those tables applied too. 

I can get behind that attribute option for mj-table as well, that makes sense too. 

Thanks!

PR #2302, opened 2021-06-20
add role="presenation" to MJML Section

mohamedsalem401:
resolves: #2294 
For accessibility, added role="presentation" to MJML Section table for Outlook.


iRyusa:
👍  Thanks

PR #2305, opened 2021-06-22
components_2.md_for_mj-include

GarryFlemings:
### components_2.md_for_mj-include

Some time ago, I opened PR #2033, about documentation for `mj-include`. I recently closed it.

In discussion, some ideas were somehow lost. This re-introduces some of those.

Some ideas in #2033 needed to come out; they're gone.

Valuable new capability for "mj-include" has since been added. Related documentation needed to stay. It did.

I found ways to change the expression in minor ways to make the text "more difficult to misunderstand", a different thing than "possible to understand".

For example, active voice is often shorter and more direct than passive voice in technical writing. I changed some without changing sentence meaning.

I changed "a" to "an" a couple of places.

"MJML", "HTML", and "CSS" are more specific and traditional (in text) than their lower-case equivalents. (Doesn't apply to code.)

I removed reference to "rendering", which referred here to MJML's compilation to HTML.

Almost all the above changes shortened the text.

I mean no offense in any changes. I could well be changing my own previous text here.

I added a little text, believing it adds clarity. You will not surprise me to consider it too wordy. I include it only in case I misread you.

### DIFF
--- a/doc/components_2.md
+++ b/doc/components_2.md
@@ -1,7 +1,8 @@
 ## mj-include
 
-The `mjml-core` package allows you to include external `mjml` files
-  to build your email template.
+The `mjml-core` package adds `mj-include` content to
+  a copy of your MJML markup in memory.
+It uses that copy to build the HTML file.
 
 ```xml
 <!-- header.mjml -->
@@ -12,10 +13,6 @@ The `mjml-core` package allows you to include external `mjml` files
 </mj-section>
 ```
 
-You can wrap your external mjml files inside the default `mjml > mj-body`
-  tags to make it easier to preview outside the main template.
-
-
 ```xml
 <!-- main.mjml -->
 <mjml>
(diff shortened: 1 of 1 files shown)


doc/components_2.md
```diff
@@ -1,7 +1,8 @@
 ## mj-include
 
-The `mjml-core` package allows you to include external `mjml` files
-  to build your email template.
+The `mjml-core` package adds `mj-include` content to
```
iRyusa, line 3: The old version is clearly better here, I don't know why would you mention "memory" nor "copy".

doc/components_2.md
```diff
@@ -25,23 +22,40 @@ You can wrap your external mjml files inside the default `mjml > mj-body`
...
   <mj-include path="./styles.css" type="css" />
   <mj-include path="./inline-styles.css" type="css" css-inline="inline" />
 ```
 
-You can also include external `html` files. They will be inserted the same way as when using a `mj-raw`.  
-You need to specify that you're including a html file using the `type="html"` attribute.  
+MJML can include external `html` files;
+  it inserts them as if finding an `mj-raw`.  
+When including an HTML file, use the `type="html"` attribute.
+MJML puts the external HTML where it finds the `<include>`.
```
iRyusa, line 58: No need to specify this as it works like that for everything.

doc/components_2.md
```diff
@@ -25,23 +22,40 @@ You can wrap your external mjml files inside the default `mjml > mj-body`
...
+  it finds the `mj-include`.
+
+MJML also allows the `mjml`, `mj-head`, and `mj-body` tags in include files.
+The resulting MJML include file can preview better outside the main template.
+When the main `mj-body` section holds these mixed include files,
+  MJML puts content of
+  (1) the `mj-head` from `mj-include` in the main `mj-head` element and 
+  (2) the `mj-body` from `mj-include` where it finds the `mj-include`.
+
+Include files can have `mj-include` elements.
```
iRyusa, line 39: I'm fine about adding a new introduction about `mj-include`, but here it's way too much text.

2 things need to be documented here : 
- Use of include file can be direct (without mjml root tag) or a complete MJML document
- Included file can use mj-include too and the path will be relative to the included file.

iRyusa (changes requested):
Can't be merged like that for now. We really need to keep the tone of the documentation instead of a colder approach, documentations like React, Angular, Styled Components are great exemples and should be a source of inspiration for further contributions.

doc/components_2.md
iRyusa: > and the path will be relative to the included file. 

Removing that part as it common sense to work like that.

Issue #2299, opened 2021-06-17
v4.10.0 is breaking our app


ddolcimascolo:
**Describe the bug**

Our build starts to fail while updating mjml to v4.10.0

**To Reproduce**

Upgrade to v4.10.0 from v4.9.3

**Expected behavior**

Our builds still pass, and emails are rendered :+1: 

**MJML environment (please complete the following information):**

 - OS: Ubuntu 20.04
 - MJML Version 4.10.0
 - MJML tool used [e.g MJML App]: The MJML API

**Additional context**

```
/builds/stordata/cmdb/node_modules/mjml-core/lib/index.js:66
var _classCallCheck2 = _interopRequireDefault(require("@babel/runtime/helpers/classCallCheck"));
                       ^
TypeError: _interopRequireDefault is not a function
    at Object.<anonymous> (/builds/stordata/cmdb/node_modules/mjml-core/lib/index.js:66:24)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Module.replacementCompile (/builds/stordata/cmdb/node_modules/append-transform/index.js:60:13)
    at Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
    at Object.<anonymous> (/builds/stordata/cmdb/node_modules/append-transform/index.js:64:4)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Module.require (internal/modules/cjs/loader.js:952:19)
    at require (internal/modules/cjs/helpers.js:88:18)
    at Object.<anonymous> (/builds/stordata/cmdb/node_modules/mjml/lib/index.js:12:41)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Module.replacementCompile (/builds/stordata/cmdb/node_modules/append-transform/index.js:60:13)
    at Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
    at Object.<anonymous> (/builds/stordata/cmdb/node_modules/append-transform/index.js:64:4)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Module.require (internal/modules/cjs/loader.js:952:19)
    at require (internal/modules/cjs/helpers.js:88:18)
    at Object.<anonymous> (/builds/stordata/cmdb/backend/mailer.js:2:390)
[...]

PR #2332, opened 2021-07-22
update @babel/runtime dep on all packages

doits:
follow up to #2299


iRyusa (approved):
👍

marklawlor:
@iRyusa Can this please be published, v4.10.x is still breaking on our app

dan-kez:
Hi @iRyusa apologies for the double ping but it would be excellent to publish a new version with this fix. We're running into the same issue as the original author.

PR #2370, opened 2021-09-28
style: make align option consistent

pahan35:
Describe **align** option as it is done for https://github.com/mjmlio/mjml/tree/master/packages/mjml-text

### DIFF
--- a/packages/mjml-divider/README.md
+++ b/packages/mjml-divider/README.md
@@ -20,17 +20,17 @@ Displays a horizontal divider that can be customized like a HTML border.
   </a>
 </p>
 
-| attribute                  | unit               | description                                        | default value |
-| -------------------------- | ------------------ | -------------------------------------------------- | ------------- |
-| border-color               | color              | divider color                                      | #000000       |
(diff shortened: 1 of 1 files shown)


iRyusa:
Looks like to be used here too
https://github.com/mjmlio/mjml/blob/master/packages/mjml-text/src/index.js#L50

Mh... I don't know what to do about this.

Maybe add a text-align attribute that fallback to align to ensure retro
compatibility ?

kmcb777:
Actually i checked the "allowedAttributes" in the code and the unit is `enum(left,center,right)` so the validator is literally expecting one of these 3 values. So i think it should stay like this in the "unit" column.
But it should indeed be consistent with mj-text. So we should change the Readme for mj-text's align instead, because it's the same, the unit is `enum(left,right,center,justify)`, so it should be in the unit column
@iRyusa any opinion on this ?

iRyusa:
I think we should support both `align` and `text-align` because you might want to have a `justify` for text-align in the div too

PR #2372, opened 2021-10-11
fix broder radius not working

gowram:
It was not working earlier, this fix will allow border radius will applies both at section and wrapper levels.


iRyusa:
Border-radius is only applied on the table as expected to ensure compatibility. However it doesn't work with a border if you don't use `border-collapse: separate;` on the table that could lead to some inconsistency on Outlook.

We'll keep it that way so

PR #2388, opened 2021-11-16
delete mj-image doc width 100%

GarryFlemings:
Earlier today, Kajsa Eklöf suggested the "width" line of the documentation for mj-image is confusing

width                         | px            | image width                    | 100%

(The attribute only allows px, but the default is in percent.)

This will change the line to

width                         | px            | image width                    | n/a

In case that's how you want to respond ...


kmcb777:
Hi, we should indeed change this, but i think it could be "parent width" instead of "n/a", so that it still gives a bit of information

iRyusa (approved):

Issue #2249, opened 2021-05-03
JSON .mjmlconfig does not support preprocessors


maxiruani:
**Is your feature request related to a problem? Please describe.**
I wanted to configure a set of preprocessors within the `.mjmlconfig` but due to its's JSON nature, it is not possible.

**Describe the solution you'd like**
I would be nice if the `.mjmlconfig` could be a `.js` file or have an alternative `mjml.config.js` file.

iRyusa:
It might be something we want to look at @kmcb777 ?

PR #2405, opened 2022-01-12
ref #2249 allow to use a .mjmlconfig.js, allow to provide preprocessors

kmcb777:


iRyusa:
Lint and rdy to merge

Issue #2396, opened 2021-12-03
Unable to add class to the <body> tag


mzrnsh:
I am trying to implement a [solution](https://www.hteumeuleu.com/2021/fixing-gmail-dark-mode-css-blend-modes/) for an issue related to how Gmail iOS app's dark mode works.

For this solution to work, I need the `<body>` tag of the generated markup to have a class `.body`. 

There could be another way to target the body tag within Gmail, but working on this made me realize that it's not currently possible to add a class to the `<body>` tag via MJML.

I might have missed something, so apologies if that's the case. Here's what I've tried:

1. Use `css-class` attribute on `mj-body`:

```html
<mj-body css-class="body">
```
This applies the class to the `div` nested within `body`, not to `body`directly.

2. Add class as an html attribute via `mj-html-attributes`:

```html
<mj-html-attributes>
  <mj-selector path="body">
    <mj-html-attribute name="class">body</mj-html-attribute>
  </mj-selector>
</mj-html-attributes>
```

This doesn't do anything, so it makes me think the `body` selector doesn't work. Other CSS selectors did work so I don't think I am using the `mj-html-attributes` wrong.

iRyusa:
I think there's an open issue about this somewhere. We really need to move the body tag in mj-body in next breaking version so css-class can be applied on the top level.

cc @kmcb777 Do we have some other breaking change that we can push for a v5 ?

matthewgovaere:
Running into this same issue — need to add a class directly to the `body` tag, but haven't had any luck.

Are there any temporary workarounds?

mzrnsh:
@matthewgovaere Are you using a build tool like Gulp with MJML?

If yes, you could watch the output file and replace `<body` with `<body class="yourClass"` on each change.

matthewgovaere:
Thanks @mzrnsh — that's exactly what I ended up doing. Not ideal, but it'll work for now!

jlarmstrongiv:
Doesn’t work for the `mjml` / `html` tag either.  I ended up using the replace method too.

mr-scrpt:
For the gmail and iso hack, you need to be able to set a class for the native body tag
```
      /* Gmail fix for white text in dark mode */
      u + .body .gmail-screen {
        background: #000000 !important;
        mix-blend-mode: screen !important;
      }

      u + .body .gmail-difference {
        background: #000000 !important;
        mix-blend-mode: difference !important;
      }
```
But over the years never given the opportunity to do so

[...]

Issue #2355, opened 2021-09-11
Allow for `lang` attribute in `<mj-body>`


rootwork:
**Is your feature request related to a problem? Please describe.**

From @hteumeuleu https://github.com/hteumeuleu/email-guidelines#lang:

> Defining the `lang` of the HTML content helps assistive technologies like screen readers to pick the right voice to read the content. The `lang` attribute needs to be defined with a valid [language tag](https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry) on the `<html>` opening tag. But because some email clients (especially webmails) remove the `<html>` element, the `lang` attribute also needs to be set on a wrapping element within the `<body>`.
>
> ```html
> <!DOCTYPE html>
> <html lang="en">
>   <head></head>
>   <body>
>     <div lang="en">
>     </div>
>   </body>
> </html>
> ```

MJML currently accepts a `lang` attribute on `<mjml>`, which it places on the root `<html>` element, but it does not accept one on the `<mj-body>` element, which would map to the containing `<div>` element that it creates.

**Describe the solution you'd like**

Allow for `lang` on `<mj-body>`.

**Describe alternatives you've considered**

The only current workaround that I've found is to nest a `<div>` within every `<mj-column>` and set the language there, which is a pain.

**Additional context**

This is arguably an a11y bug, but since it's adding a new enhancement I used the feature issue template.

I'd be happy to do a PR for this if the maintainers agree with the idea.

GarryFlemings:
### Alternative

The `lang` attribute on `<mjml>` also adds the `lang` annotation on the `div` after `<mj-body>`. MJML has no `lang` attribute on `<mj-body>`.

### Comments

Good suggestion! And, taking nothing away from it ...

This option is appropriate if no user would want, for example, `lang=en` in one place and `lang=fr` in the other. Maybe that's expectable.

This option abstracts a little complexity from MJML users, an added value for MJML.

This option makes MJML more maintainable because the team is less likely to get questions about the difference between the two attributes.

This option keeps MJML more consistent with HTML. MDN documents the `lang` attribute for `<html>` (our equivalent being `mjml`), but not for `<body>` (ours is `mj-body`).

iRyusa:
👍 I think we should push `globalDatas` (and probably rename this with a smarter name) into bodyHelpers here https://github.com/mjmlio/mjml/blob/master/packages/mjml-core/src/index.js#L268 

[...]

PR #2408, opened 2022-01-13
ref #2396 #2355 move body tag from skeleton to mj-body, add lang on w…

kmcb777:
…rapping div


iRyusa:
- Add `id` attribute to `body` for some fixes (slack comment)

iRyusa:
I think we should just merge it and not bump into a major version. This is not really that big of a breaking changes as it can only mess with few selector for some users

dazgreer:
Updated in v5.x

PR #2422, opened 2022-02-04
Add text-align="center" to doc example

cmacdonnacha:
Just 2 quick changes:

- Added text-align="center" to doc example since the [image](https://documentation.mjml.io/#basic-layout-example) shows it centered and the [doc itself](https://documentation.mjml.io/#company-header) says  "The first section of the email consists in a centered banner".

- Fixed some formatting  directly below

The [Live example](https://mjml.io/try-it-live/P16bj5WQT) will also have to be updated but I'm not sure how to go about changing that.


iRyusa:
👍  thanks

PR #2442, opened 2022-03-14
Fix overlapping max-width and min-width on media queries

Wanztwurst:
The configured breakpoint is being used in the css media queries for the desktop `(min-width: breakpoint)` and mobile `(max-width: breakpoint)`. This causes the queries to overlap at the exact breakpoint size, e.g. 700px, and being active at the same time.

These changes reduce the mobile breakpoint to `(max-width: breakpoint-1)`, so that the set breakpoint represents the smallest width for the desktop layout.

### DIFF
--- a/packages/mjml-head-breakpoint/README.md
+++ b/packages/mjml-head-breakpoint/README.md
@@ -1,5 +1,5 @@
 ## mj-breakpoint
-This tag allows you to control on which breakpoint the layout should go desktop/mobile.
+This tag allows you to control on which breakpoint the layout should go desktop/mobile. At window widths below this value the layout will be displayed in mobile mode.
 
 ```xml
 <mjml>
--- a/packages/mjml-image/src/index.js
+++ b/packages/mjml-image/src/index.js
@@ -1,6 +1,6 @@
 import { min } from 'lodash'
 
-import { BodyComponent } from 'mjml-core'
+import { BodyComponent, makeLowerBreakpoint } from 'mjml-core'
 
 import widthParser from 'mjml-core/lib/helpers/widthParser'
 
@@ -133,7 +133,7 @@ export default class MjImage extends BodyComponent {
   }
 
   headStyle = (breakpoint) => `
-    @media only screen and (max-width:${breakpoint}) {
+    @media only screen and (max-width:${makeLowerBreakpoint(breakpoint)}) {
       table.mj-full-width-mobile { width: 100% !important; }
       td.mj-full-width-mobile { width: auto !important; }
     }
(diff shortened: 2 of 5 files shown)


iRyusa:
I don't really get your issue as it was clearly intended to have those media queries to overlap here. Can you add some context ?

Wanztwurst:
I think it was made with the same value for min- and max-width out of convenience. Why would the rules for desktop and mobile apply both at the breakpoint? It's either desktop or mobile. The breakpoint being the width the mailing is created for I don't see why objects would already be supposed be displayed differently.

iRyusa:
Even if MJML is mobile first, both case of mj navbar and full width image media queries are targeted only for mobile. There’s the closest way to have a progressive enhancement without compromising desktop or mobile version.

iRyusa:
Well I'm stupid.
As both media query are turned on at 480px (default breakpoint) it means that there's a mix of both version eg full width image is applied at 480px on desktop version. So that's a great change 👍

packages/mjml-head-breakpoint/README.md
```diff
@@ -1,5 +1,5 @@
 ## mj-breakpoint
-This tag allows you to control on which breakpoint the layout should go desktop/mobile.
+This tag allows you to control on which breakpoint the layout should go desktop/mobile. At window widths below this value the layout will be displayed in mobile mode.
```
iRyusa, line 2: I'm not sure we should add this in the doc here as everyone should be familiar enough with media queries and what a breakpoint mean. Can you remove this ?

packages/mjml-image/src/index.js
```diff
@@ -133,7 +133,7 @@ export default class MjImage extends BodyComponent {
   }
 
   headStyle = (breakpoint) => `
```
iRyusa, line 135: Maybe we could pass once for all a mobileBreakpoint and desktopBreakpoint ? It changes a bit the API but at least it will be computed only once ?

packages/mjml-head-breakpoint/README.md
Wanztwurst: Sure, if it's too verbose.

packages/mjml-image/src/index.js
Wanztwurst: I'm unsure because of the naming, but I'll do this.

iRyusa: Yeah might be better to have an object instead of only  breakpoint so it would be extensible in the future.
I don't know if a MJML5 release is near soon so :/

PR #2462, opened 2022-04-20
Add protocol to parcel link

georgeblahblah:
* Adds a protocol (`https://`) before the parcel link, otherwise it's treated as relative


avigoldman:
Great catch 🙏

Issue #2144, opened 2021-01-15
Centering a group of dynamic width columns 


kprokopczyk:
Hello, i have 3 columns that i want to smash together and center horizontally _as a group_. In other words, I want the image, slash and text to appear as **one object** centered horizontally, so something like this:
![image](https://user-images.githubusercontent.com/30055821/104780374-4f053280-5746-11eb-8538-64c4e02e8471.png)

I can't figure out how to do it without a `width="auto"` option on `mj-column`. Any pointers? Is a 3 column layout the right approach or is there a better solution?
```
<mjml>
	<mj-head>
		<mj-style inline="inline">
			.slash {
				padding-right: 6px;
				font-size: 20px;
				font-style: normal;
				color: blue;
			}
			div.groupCenter {
				text-align: center !important;
			}
		</mj-style>
	</mj-head>
	<mj-body>
		<mj-section>
			<mj-group css-class="groupCenter">
				<mj-column width="120px" vertical-align="middle"  border="1px solid black">
					<mj-image padding="0" align="right" src="https://logosbynick.com/wp-content/uploads/2018/03/final-logo-example.png"/>
				</mj-column>
				<mj-column vertical-align="middle" padding="10px 0px"  border="1px solid red">
					<mj-text align="center">
						<span  class="slash">&#8725;</span>
					</mj-text>
				</mj-column>
				<mj-column vertical-align="middle" border="1px solid black">
					<mj-text align="left" padding='0' font-weight='200' font-size='16px'>
						Example
					</mj-text>
				</mj-column>
			</mj-group>
		</mj-section>
	</mj-body>
</mjml>
```

iRyusa:
"auto" width doesn't exist because it need to generate outlook ghost table.

Inside mj-groupe you can only define width in `%` and not in `px`, you should just do something like 40/20/40 inside this mj-groupe and it will works fine.

kprokopczyk:
Thank you, but I left out one important element: The third column has a variable width.

iRyusa:
There's no way to have "auto" column in MJML because of ghost table. So you should just play with width size to handle most cases.

kprokopczyk:
Alright, thanks for the help.

caseyjhol:
[...]

PR #2468, opened 2022-04-29
Support width="auto" on mj-column

caseyjhol:
Addresses #2144 

Code used to test:

```
<mjml>
  <mj-body>
    <mj-section>
      <mj-column width="250px" background-color="blue">
        <mj-button>250px</mj-button>
      </mj-column>
      <mj-column width="auto" background-color="red">
        <mj-button>auto</mj-button>
      </mj-column>
      <mj-column width="auto" background-color="yellow">
        <mj-image src="https://mjml.io/assets/img/logo-small.png" />
      </mj-column>
    </mj-section>
  </mj-body>
</mjml>
```

Tested in:
Outlook 365
Outlook 2007
image
Gmail
image
Yahoo webmail
image
Outlook webmail
image

### DIFF
--- a/packages/mjml-column/src/index.js
+++ b/packages/mjml-column/src/index.js
@@ -27,7 +27,7 @@ export default class MjColumn extends BodyComponent {
     'inner-border-top': 'string',
     padding: 'unit(px,%){1,4}',
     'vertical-align': 'enum(top,bottom,middle)',
-    width: 'unit(px,%)',
+    width: 'unit(px,%,auto)',
   }
 
   static defaultAttributes = {
@@ -122,6 +122,9 @@ export default class MjColumn extends BodyComponent {
     const width = this.getAttribute('width')
     const mobileWidth = this.getAttribute('mobileWidth')
 
+    if (width === 'auto') {
+      return width
+    }
     if (mobileWidth !== 'mobileWidth') {
       return '100%'
     }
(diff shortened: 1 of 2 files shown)


caseyjhol:
@iRyusa Thoughts?

iRyusa:
I have COVID so I'll try to play a bit with it when I'm back. I have few case in mind that we should test

iRyusa:
Ok so I've played with it a bit and here a real use-case that doesn't work : 
```
<mjml>
  <mj-body>
    <mj-section>
      <mj-column width="150px" background-color="blue">
        <mj-button>250px</mj-button>
      </mj-column>
      <mj-column width="auto" background-color="yellow">
        <mj-image src="http://via.placeholder.com/640x360" />
      </mj-column>
      <mj-column width="100px" background-color="red">
        <mj-button>auto</mj-button>
      </mj-column>
    </mj-section>
  </mj-body>
</mjml>
```

Can't share the litmus test but it overflow on almost every clients. I don't know if there's a way to prevent that and if other component could just overflow the max-width too. It's not really a safe option (adding potential issue and confusion to user+ extra support on slack)

caseyjhol:
For reference, this is the above example:

Chrome:
image

Outlook 2007:
image

I'll do some more investigating.

iRyusa:
Closing for now as it's stale feel free to reopen if you find some way to prevent this.

Issue #470, opened 2016-12-09
Vertical Align image inside <mj-column>


jaypeedevlin:
I'm trying to vertically align an image inside an `<mj-column>` tag, but can't seem to find away.  The `vertical-align` parameter seems to do nothing so must not affect the content within it.

My feeling is that this is not supported, are you able to confirm if this is correct?

Simplified code that I have so far

```
<mjml>
  <mj-body>
    <mj-container>
      <mj-section>
        <mj-column width="25%" vertical-align="middle">
          <mj-image align="center" src="http://placehold.it/100" />
        </mj-column>
        <mj-column width="75%">
          <mj-text>
          <h4>I want the image on the left to be vertically centered</h4>
          <p>Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. 
          </p>
          <p>Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. 
          </p>
          </mj-text>
        </mj-column>
      </mj-section>
    </mj-container>
  </mj-body>
</mjml>
```

and how this renders:

![image](https://cloud.githubusercontent.com/assets/17235807/21057051/b2070e1c-bdfd-11e6-9472-864ef812ffee.png)

ngarnier:
Hey @jaypeedevlin, thanks for reporting the issue! There's definitely an issue with `vertical-align` here, we'll investigate. Interestingly, it does work if set to `top` or `bottom`, and it works if `vertical-align` is set to `middle` on both columns ([live example](https://mjml.io/try-it-live/rydItYdQl)):
```
<mjml>
  <mj-body>
    <mj-container>
      <mj-section>
        <mj-column width="25%" vertical-align="middle">
          <mj-image align="center" src="http://placehold.it/100"  />
        </mj-column>
        <mj-column width="75%" vertical-align="middle">
          <mj-text>
          <h4>I want the image on the left to be vertically centered</h4>
[...]

PR #2472, opened 2022-05-12
Update docs with note about vertical align bug

altescape:
Note about issue https://github.com/mjmlio/mjml/issues/470 which is closed but not yet fixed as far as I can tell.


iRyusa:
It's a small gotcha of HTML vertical-align but why not add this to the doc 👍

Issue #1690, opened 2019-08-26
Random `carouselId` breaks Jest-Snapshot tests


12finger:
Concerns [mjml-carousel](https://github.com/mjmlio/mjml/tree/master/packages/mjml-carousel) only.

This line](https://github.com/mjmlio/mjml/blob/master/packages/mjml-carousel/src/Carousel.js#L45) generates a fully random id which is used on many css-classes and elements. On each render run it is different, so all snapshot tests involving a file that renders a carousel, brake.

<img width="1792" alt="notification-service-template-engine 2019-08-26 23-49-10" src="https://user-images.githubusercontent.com/1730022/63725980-39931e00-c85c-11e9-85be-2c1ae962257d.png">

Would be nice to have a static version for testing ?!
Or actually why use that random id in the firstplace ?

iRyusa:
Well the id is needed to ensure that you can use multiple carousel in a single email. You can just stub the crypto package for that https://github.com/mjmlio/mjml/blob/master/packages/mjml-carousel/src/Carousel.js#L45

12finger:
@iRyusa  thanks so much for the super quick reply. 
and: True! 🤦‍♂ stubbing is the way to go ..

PR #2478, opened 2022-06-06
Removed the 'crypto' dependency

3bl3gamer:
The `crypto` module brings a *huge* amount of code to the browser build and it's removal reduces the build size by **400 KiB** or by 25%.

The `crypto.randomBytes` [Generates cryptographically strong pseudorandom data](https://nodejs.org/api/crypto.html#cryptorandombytessize-callback). That strength is not needed since these hex strings are used only as ids/classes in the generated HTML.

This change may cause test fails [for some users](https://github.com/mjmlio/mjml/issues/1690), but this can be fixed by stubbing `Math.random`.

Another option would be to generate sequential IDs:

```js
let lastId = -1
export default function genStringId() {
  lastId += 1
  return lastId+''
}
```

This won't cause problems with tests but may reveal the amount of e-mails generated by some application (if used on the server side).


3bl3gamer:
`BundleAnalyzerPlugin` for comparison. Before:

before

After:

after

iRyusa (approved):

iRyusa:
I'll try to get this in 4.13.1

iRyusa:
@3bl3gamer Can you solvle those conflict i'd like to release this next week ?

3bl3gamer:
@iRyusa sorry for the delay, resolved.

F21:
Any change this can be merged soon?

Issue #1757, opened 2019-11-18
Confused about purpose of mjml-text


callumlocke:
Hi. I am just learning MJML. The [docs for `mjml-text`](https://mjml.io/documentation/#mjml-text) suggest you can nest "any HTML tag with any attributes". And it gives this example:

```html
<mjml>
  <mj-body>
    <mj-section>
      <mj-column>
        <mj-text>
          <h1>
            Hey Title!
          </h1>
        </mj-text>
      </mj-column>
    </mj-section>
  </mj-body>
</mjml>
```

The description is simply:

> This tag allows you to display text in your email.

But what does it _do_? I can see that it lets me specify text styles, but then I could also do that on the `h1` directly. In the case of `font-size`, I would have to add it on the `h1` either way, because otherwise the browser default `h1` font-size will override whatever font-size I set on the `mj-text`... so, why use the `mj-text` at all? What's it for?

It could do with a bit more explanation than "This tag allows you to display text in your email." :)

I tried the Getting Started tutorial but it doesn't shed much light.

iRyusa:
It’s a standard tag for content inside a column. Mj-text also has extra attributes as container background color, padding. You can also style them with mj attributes.

ngarnier:
Basically `mj-text` enables you to add Text inside a column. The fact it supports HTML is simply for more flexibility but that doesn't make the usage of `mj-text` attributes useless. 

Consider this simple example: https://mjml.io/try-it-live/SJRh3wlhS

Simply put, it enables to:
- Style all text children at once for common styles (here with `color`)
- Reuse HTML semantics for structure and styling (here with `h1`/`p`)
- Override some children like you would in HTML / CSS (here with `span` that has its own `font-family`)

And all this without having to create a different `mj-text` each time.

jglovier-okta:
>Consider this simple example: https://mjml.io/try-it-live/SJRh3wlhS
Ahhh this is super helpful. Thank you!!

This had me very confused as well. I didn't realize that the `<mj-text>` component compiles to `<div>` tags. 😄 I've been using MJML to update legacy email templates for the past week and thought that `<mj-text>` just rendered as `<p>` tags and so I was just limited to styling `<p>` tags as headings. I just went searching for a way to add heading tags in MJML and discovered this issue so now I understand it's just a `<div>` wrapper that I can put any content inside. Thanks!

PS - it might be helpful to see that a bit more explicitly in the docs. 🙏

PR #2492, opened 2022-06-30
improve mj-text doc

mifi:
from https://github.com/mjmlio/mjml/issues/1757

### DIFF
--- a/packages/mjml-text/README.md
+++ b/packages/mjml-text/README.md
@@ -1,16 +1,17 @@
 ## mj-text
 
-This tag allows you to display text in your email.
+This tag allows you to display text and HTML in your email and has shorthand attributes for easily styling its children.
 
  ```xml
 <mjml>
   <mj-body>
     <mj-section>
       <mj-column>
-        <mj-text>
-          <h1>
-            Hey Title!
-          </h1>
+        <mj-text font-family="Helvetica" color="#F45E43">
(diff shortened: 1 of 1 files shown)


packages/mjml-text/README.md
```diff
@@ -19,7 +20,7 @@ This tag allows you to display text in your email.
  ```
 
 <p style="text-align: center;" >
-  <a href="https://mjml.io/try-it-live/components/text">
+  <a href="https://mjml.io/try-it-live/E7Tk_8zjI">
```
iRyusa, line 23: You don't need to change the link here it will fetch the content automatically 👍

packages/mjml-text/README.md
```diff
@@ -1,16 +1,17 @@
 ## mj-text
 
-This tag allows you to display text in your email.
+This tag allows you to display text and HTML in your email and has shorthand attributes for easily styling its children.
```
iRyusa, line 3: > and has shorthand attributes for easily styling its children

Not sure this provide any value here. But 👍 to add HTML mentioned

mifi: not sure what is the purpose of `font-family="Helvetica" color="#F45E43"` etc other than being shorthands

iRyusa: I think this is self-explanatory, as you can say this for any component

PR #2500, opened 2022-07-08
[POC] Try to override padding into margin when background for Outlook

iRyusa:


iRyusa:
Blocked for now, need to apply margin on elements that supports it. But in some component it won't behave the same with background color as background is applied in the box not around the margin

PR #2510, opened 2022-07-31
Add list of ports and wrappers to the documentation

SebastianStehle:


iRyusa:
I'm OK to add this but we really need to add a big warning that MJML
maintainers are not involved in those projects.
They might miss some key features listed in this documentation, so we
provide 0 support on those.

SebastianStehle:
I have added a warning and also added a few hints about the limitation of the ports.

Issue #2556, opened 2022-10-03
Output pure mjml after processing includes, for easy copy paste into try-it-live


martyzz1:
**Is your feature request related to a problem? Please describe.**
I'd like to be able to test the mjml from my project in the try-it-live editor, without having to manually copy and paste individual "includes" into the editor piecemeal.

**Describe the solution you'd like**
add an option to the mjml cli call to say output-format=mjml

**Describe alternatives you've considered**
thought about doing a bash script, but it would be rubbish and clunky.

nbouvrette:
I've been playing with `mjml` for a few days now and I am seeing a lot of challenges around tooling. For example, there is a decent amount of dependency on the online tool to make this solution viable.

An alternative would be to send emails every time you make a modification which would be quite inefficient.

Now, since the tool doesn't come with templating capabilities, what happens when you start adding templating markup in your template... well the live editor reports all sorts of markup errors. At least this is what happens when you use EJS and I bet it would be the same if you use anything other than handlebars.

I'm still trying to wrap my head around what are good functionalities VS the ones you should avoid.. including files, for now, have been something I was trying to avoid for the reasons you are mentioning.

iRyusa:
We have everything to do so, someone need to open a PR on this
https://github.com/mjmlio/mjml/issues/2465#issuecomment-1109515536

nbouvrette:
@iRyusa it looks like this would solve the problem to put the content in the editor but did you think about how to merge back the edits into the individual files?

iRyusa:
There’s no way to do so easily.On 4 Oct 2022, at 19:07, Nicolas Bouvrette ***@***.***> wrote:﻿
@iRyusa it looks like this would solve the problem to put the content in the editor but did you think about how to merge back the edits into the individual files?

nbouvrette:
I agree... I was thinking of either:

- "[codesandbox](https://codesandbox.io/)"-alike editor where you have some sort of download option to get back your files. This would be a lot of work and possibly tricky to reproduce local paths to create packages...
- VSCode plugin where you can see the preview directly in VSCode, using your local environment - seems like a good option for developers. Not sure how much work it would be and this probably wouldn't be aligned with your (Mailjet) commercial goals for developing this tool.

iRyusa:
I'm not affiliated with Mailjet. Both project would require too much time
and effort.
Let's not deviate too much from original issue

nbouvrette:
[...]

Issue #914, opened 2017-12-10
renderMJML children


maxs15:
I'm trying MJML 4 and I can't render the children of my custom component.
Here is what I'm doing, am I doing something wrong ?
Sorry if I'm missing something, always hard without doc ;) Thanks for the help.

Custom component
```js
class Layout extends BodyComponent {

	render() {
		return this.renderMJML(`
			<mj-section>
				<mj-column>
					${this.renderChildren(this.props.children)}
				</mj-column>
			</mj-section>
		`)
	}

}
```

mjml template:
```js
<mjml>
	<mj-body>
		<layout>
			<mj-text>TEST</mj-text>
		</layout>
	</mj-body>
</mjml>
```

iRyusa:
Hi @maxs15 

I think this is because mj-column can't receive plain HTML ( this.props.children will already be rendered when passing it into the mj-column ). I think this is something we didn't had in mind when creating this feature. We might need some kind of rawChildren that would allow this kind of thing.

maxs15:
Hey @iRyusa,

Ok I see what's the issue, rawChildren would be a handy property indeed.
I hope you will consider added it in the next release, there's unfortunately lot of things that I can't do without that.

Thanks, and good work on MJML 4 🎉

iRyusa:
ATM you can actually convert your JSON as XML with this snippet : https://jsfiddle.net/jp1nfsk0/ I think it should work fine 👍 Can you try it ?

maxs15:
Awesome it works ! I'm gonna use that for the time being 👍 
You could actually expose such a method in the BodyComponent class (renderRawChildren for instance)

iRyusa:
Yeah I think it will either be exposed as an helper method or as you said on BodyComponent, we'll talk about this 👍

iRyusa:
@kmcb777 Can you add an option on`renderChildren` func to get the raw XML version and allow this for the next beta ?

PR #2558, opened 2022-10-04
Expose raw xml api

Asherlc:
Right now, when rendering raw XML, there's no easy way to override it on a component basis, in the same way we allow `render` to be overriden when rendering HTML. This PR moves XML stringifying logic from the `jsonToXML` function, to a method called `renderAsMJML` (I found the idea of `rawXML` a bit unclear -- both MJML and HTML are raw xml, right?).

Example use case is escaping HTML or quotation marks etc in attributes.


iRyusa:
Well it's a very old option that make no real sens in naming, I totally agree with you on that (context: #914).

Maybe we should rename this option at first and deprecate the `rawXML`.

packages/mjml-core/src/helpers/jsonToXML.js
```diff
@@ -1,16 +0,0 @@
-const jsonToXML = ({ tagName, attributes, children, content }) => {
```
iRyusa, line 1: We really need to keep that as a helper function as I know some tools use this API. And we might as well need it for #2556 to resolve include into an outputfile.

Asherlc: no problem, i've restored it and modified `createComponent` to leverage it

packages/mjml-core/src/createComponent.js
```diff
@@ -196,6 +195,18 @@ export class BodyComponent extends Component {
     )
   }
 
+  renderAsMJML() {
```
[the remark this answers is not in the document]
Asherlc: sure thing, i've renamed this and the XML-related names

iRyusa:
Really busy atm but as soon as I've some time I'll review this

packages/mjml-core/src/helpers/jsonToXML.js
iRyusa, line 1: Can we keep the file name, export the helpler from the new file and throw a warning to import jsonToMJML instead of this file ?

Thanks :D

PR #2650, opened 2023-03-16
Fix % mobile width in mjml-column

martinboksa:
`getMobileWidth` in `mjml-column` should return normal percents instead of 0.xy%


iRyusa:
Thanks for this ! 👍

PR #2666, opened 2023-04-05
[WIP] Remove ref to html-minifier/js-beautify/migrate

iRyusa:
Not tested yet.

- Need to write doc about how to keep the feature available for users.
- Check how to handle this case in mjml-browser

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -19,14 +19,14 @@ A clear and concise description of what you expected to happen.
 
 **MJML environment (please complete the following information):**
  - OS: [e.g. MacOS]
- - MJML Version [e.g. 4.2.0]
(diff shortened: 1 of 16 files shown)


phlogisticfugu:
@iRyusa - is this PR still the direction you want to go?

we want to use mjml, but currently cannot due to vulnerabilities in html-minifier and js-beautify

if you are open to PRs, could help with removing them

also, see in https://github.com/mjmlio/mjml/commits/v5.0.0-alpha.1/ that you more recently were thinking of replacing them.

iRyusa:
Hi @phlogisticfugu ,
Minifier seem to be mandatory for MJML as it can prevent some bugs in some email clients, so yeah removing doesn't seem the best solution in our case.
For js-beautify, it's a different story as it make the "raw" output from MJML somewhat readable. I don't think this is a mandatory feature, but I think it's better to have a properly formatted output.

PR #2669, opened 2023-04-07
Update cheerio and htmlparser2

### LINKED ISSUE #2668: Please update `cheerio` to latest
`mjml-core@4.14.0` uses `cheerio@1.0.0-rc.10` and `juice@9` uses `cheerio@1.0.0-rc.12` (latest). So we currently have many package duplications with different versions in node_modules:
- dom-serializer
- domhandler
- htmlparser2
- parse5
- parse5-htmlparser2-tree-adapter
- css-select
- domutils
- ...

Semigradsky:
Fixed #2668

### DIFF
--- a/packages/mjml-core/package.json
+++ b/packages/mjml-core/package.json
@@ -23,8 +23,8 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.14.6",
-    "cheerio": "1.0.0-rc.10",
-    "detect-node": "2.0.4",
+    "cheerio": "1.0.0-rc.12",
+    "detect-node": "^2.0.4",
     "html-minifier": "^4.0.0",
     "js-beautify": "^1.6.14",
     "juice": "^9.0.0",
--- a/packages/mjml-core/src/index.js
+++ b/packages/mjml-core/src/index.js
@@ -14,7 +14,7 @@ import path from 'path'
 import juice from 'juice'
 import { html as htmlBeautify } from 'js-beautify'
 import { minify as htmlMinify } from 'html-minifier'
-import cheerio from 'cheerio'
+import { load } from 'cheerio'
 
 import MJMLParser from 'mjml-parser-xml'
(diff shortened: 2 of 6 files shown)


packages/mjml-core/src/index.js
```diff
@@ -14,7 +14,7 @@ import path from 'path'
 import juice from 'juice'
 import { html as htmlBeautify } from 'js-beautify'
 import { minify as htmlMinify } from 'html-minifier'
-import cheerio from 'cheerio'
```
Semigradsky, line 17: Default export was deprecated so I changed it to import specific function.

packages/mjml-core/src/index.js
```diff
@@ -14,7 +14,7 @@ import path from 'path'
...
+import { load } from 'cheerio'
```
Semigradsky, line 17: Default export was deprecated so I changed it to import specific function.

packages/mjml-parser-xml/package.json
```diff
@@ -24,7 +24,7 @@
   "dependencies": {
     "@babel/runtime": "^7.14.6",
     "detect-node": "2.0.4",
-    "htmlparser2": "^4.1.0",
+    "htmlparser2": "^8.0.1",
```
Semigradsky, line 27: `cheerio` uses `htmlparser2@8.0.1` so I changed it here to have same version

packages/mjml-core/package.json
```diff
@@ -23,8 +23,8 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.14.6",
-    "cheerio": "1.0.0-rc.10",
-    "detect-node": "2.0.4",
+    "cheerio": "1.0.0-rc.12",
+    "detect-node": "^2.0.4",
```
Semigradsky, line 27: It looks like a typo, I don't see the reason why have a specific version here.

PR #2677, opened 2023-04-26
fix(mjml-core): add print to media queries

jimmyfortinx:
**Context:**
We want to be able to use mjml to generate our invoice/statement and print them as PDF. We already have an wysiwyg editor that generates mjml for our emails and we would like to reuse it to have an editor over our invoice/statement as well.

----------------

I know the current state of my PR is not enough to have it approved and merged. I already have talked with @iRyusa in Slack and he shared his concerns around breaking email clients. I would need your help to get this PR fully tested, I'm not asking you to do it for us, but I would need some guidance.

- Could we use someone's account to get all screenshots rendered through Litmus or Email On Acid? (since we don't own any at this moment).
- Do you already have one or multiple mjml that we could use that would ensure we are testing all your components and not introducing breaking changes?

Once those questions will be answered, we will be pleased to make all the tests required to get this PR approved.

Thanks in advance for the people that will be involved in that PR.

-------------------

When the option `printerSupport` is not provided it does not introduce any breaking changes
```
const finalMjml = mjml2html(mjmlWithVariables);
```

But when you want to fix the layout to support printers you just have to add the option like that:
```
const finalMjml = mjml2html(mjmlWithVariables, { printerSupport: true });
```

PR for Typescript: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/65463

_Result in Chrome using plain html:_
image

_Result when printing it as a PDF:_
| Before | After |
| --- | --- |
| image | image |
| [before-fix.pdf](https://github.com/mjmlio/mjml/files/11333383/before-fix.pdf) | [after-fix.pdf](https://github.com/mjmlio/mjml/files/11333386/after-fix.pdf) |


packages/mjml-core/src/helpers/mediaQueries.js
```diff
@@ -18,7 +18,7 @@ export default function buildMediaQueriesTags(breakpoint, mediaQueries = {}, for
 
   return `
     <style type="text/css">
-      @media only screen and (min-width:${breakpoint}) {
+      @media only screen and (min-width:${breakpoint}), print {
```
iRyusa, line 21: Maybe we can isolate them inside it's own style to ensure it won't break anything and add an attribute on `mjml` tag to disable it by default and only add them if needed like owa media queriees

jimmyfortinx: I like the idea of disabling it by default. I'll try to figure out a way to achieve your suggestion.

jimmyfortinx: @iRyusa I applied what you've suggested and it worked perfectly. Like me know if it needs modification and what would the next steps.

iRyusa (approved):
LGTM we should be able to merge this soon

PR #2703, opened 2023-07-06
fix(mj-carousel): correct media query

jdrouet:
That media query should have a comma between screen and yahoo to be valid

### DIFF
--- a/packages/mjml-carousel/src/Carousel.js
+++ b/packages/mjml-carousel/src/Carousel.js
@@ -178,7 +178,7 @@ export default class MjCarousel extends BodyComponent {
 
       [owa] .mj-carousel-thumbnail { display: none !important; }
       
-      @media screen yahoo {
+      @media screen, yahoo {
           .mj-carousel-${this.carouselId}-icons-cell,
           .mj-carousel-previous-icons,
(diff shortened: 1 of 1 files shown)


iRyusa:
I'm not sure about this https://www.htmlemailcheck.com/knowledge-base/target-yahoo-mail-using-css-media-query/ 🤔

jdrouet:
If we follow [this documentation](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_media_queries/Using_media_queries#combining_multiple_types_or_features), to combine media queries, we need commas or `and` keyword. In our case, it should be `@media screen and yahoo` or `@media screen, yahoo` to be valid.

iRyusa:
Most client alter the HTML+CSS at the end so that's why you will find invalid stuff.
You can find multiple source about this like here https://www.emailonacid.com/blog/article/email-development/the-new-yahoo-mail-and-how-to-target-it/  + https://www.emailonacid.com/blog/article/industry-news/yahoo_mail_now_supports_media_queries/

jdrouet:
After doing some research: https://litmus.com/community/discussions/7103-targeting-for-yahoo-mail

PR #2704, opened 2023-07-06
fix(mjml-core): close negative condition

jdrouet:
According to https://en.wikipedia.org/wiki/Conditional_comment we need that square bracket to close that conditional tag


iRyusa:
👍

Issue #2695, opened 2023-06-22
Rendering another MJML template in MJML template


returnv01d:
**Describe the bug**
When MJML template contains in body another full MJML template (another <mjml>....</mjml>) then it does render only outer template (see screenshot) without throwing any errors.

**To Reproduce**
1. Create a file with this MJML code: 
```
<mjml>
  <mj-body background-color="#d2f1f6">
    <mj-section>
      <mj-column>
        <mj-text color="#F45E43">Hello World</mj-text>

      </mj-column>
    </mj-section>
    
    <mjml>
      <mj-body background-color="#F45E43">
        <mj-section>
          <mj-column>
            <mj-text>Test</mj-text>
          </mj-column>
        </mj-section>
      </mj-body>
    </mjml>
  </mj-body>
</mjml>
```
2. Render it to HTML by doing 'mjml -r <path_to_file> --config.validationLevel strict`

**Expected behavior**
There should be at least error about illegal nesting or rendering both templates.

**MJML environment (please complete the following information):**
 - OS: MacOS
 - MJML Version 4.5.1
 - MJML tool used npm MJML

**Screenshots**
<img width="1439" alt="image" src="https://github.com/mjmlio/mjml/assets/32848266/a7c3bca3-40f7-4d76-890d-3a7d36eb72fd">

**Additional context**
Embedding MJML in layout uncovered this issue.

iRyusa:
Why would you expect this to work ?

returnv01d:
I made a fix for this issue: https://github.com/mjmlio/mjml/pull/2709

PR #2709, opened 2023-07-13
fix(mjml-core): proper validation of mjml tag

returnv01d:
Fixes https://github.com/mjmlio/mjml/issues/2695
So, the dependency rules were not working correctly because `<mjml>` tag was entirely ignored. Without it MJML wasn't working, so I made new component that represents the `<mjml>` tag , so dependency rules can work properly. It can encapsulate document, for example it can be used to generate skeleton code here in render instead of special behavior for that, but it needs more refactoring then.
Couldn't find place in package to write tests so here is proof of validity:
https://github.com/mjmlio/mjml/assets/32848266/39905889-2f15-4500-b405-4edd02e91110

### DIFF
--- a/packages/mjml-document/src/index.js
+++ b/packages/mjml-document/src/index.js
@@ -1,0 +1,11 @@
+import { BodyComponent } from 'mjml-core'
+
+export default class MjDocument extends BodyComponent {
+  static componentName = 'mjml'
+
+  static endingTag = false
+
+  render() {
+    return ''
+  }
+}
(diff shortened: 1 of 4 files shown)


packages/mjml-document/src/index.js
```diff
@@ -0,0 +1,11 @@
...
+
+export default class MjDocument extends BodyComponent {
+  static componentName = 'mjml'
+
+  static endingTag = false
+
+  render() {
+    return ''
+  }
+}
```
iRyusa, line 11: I'm not really happy with this solution here, body component expect to render something, it's just a hack at the end to make it work :/ 

I guess we could add something like a "mock-class" inside MJML just here to list `mjml` attributes + dependencies but it doesn't really need to have it's own package and should live inside mjml-core

iRyusa:
Closing for now we won't go in this direction I guess.

PR #2793, opened 2023-12-06
Automatic component dependencies registration

### LINKED ISSUE #2784: Automatic component dependencies registration when dependency static key is present
### Is your feature request related to a problem? Please describe.

I'm currently working on a custom component and I struggled a bit to understand why, despite registering my component and have the proper static `dependencies` key in my class, I was forced to manually call the `registerDependencies()` helper from `mjml-validator`.  

Is this an intended behavior ? 

I guess it is worth asking because I apparently was not the first person to encounter [this problem](https://github.com/mjmlio/mjml/issues/2755).

### Describe the solution you'd like

Is it possible to check and automatically register the dependencies if the static `dependencies` key is present in the component class when using `registerComponent()` from `mjml-core` ?

Could even be behind a flag given as a second `options` argument if we want to be sure we do not break existing code.

If you think it could be a good thing to implement I'm willing to make the corresponding PR to add this feature.

### Describe alternatives you've considered

Keep on using `registerDependencies()` or `.mjmlconfig` file to register my component.

### Additional context

The component I'm working on: https://github.com/Freezystem/mjml-bar-chart

Freezystem:
### Automatic component dependencies registration
To avoid recurrent questionning when using `registerComponent` function, I added a second parameter to the function allowing the user to pass `options`.  

If the option `registerDependencies` is _truthy_, and the current component has a `dependencies` property, the component dependencies will be automatically registered.

closes #2784

### DIFF
--- a/packages/mjml-core/src/components.js
+++ b/packages/mjml-core/src/components.js
@@ -1,4 +1,5 @@
 import { kebabCase } from 'lodash'
+import { registerDependencies } from "mjml-validator";
 
 const components = {}
 
@@ -8,8 +9,12 @@ export function assignComponents(target, source) {
   }
 }
 
-export function registerComponent(Component) {
+export function registerComponent(Component, options = {}) {
   assignComponents(components, [Component])
+
+  if (Component.dependencies && options.registerDependencies) {
+    registerDependencies(Component.dependencies)
+  }
 }
 
 export default components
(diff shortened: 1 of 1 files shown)


iRyusa:
Can you save via Prettier please to make the CI happy ?

Issue #2741, opened 2023-09-08
Border radius isn't getting applied to border style


arpitjp:
**Describe the bug**
For `mj-section` & `mj-column`, border radius doesn't get applied to the border style leading to peculiar behaviour.

<img width="1200" alt="image" src="https://github.com/mjmlio/mjml/assets/37728770/52736a67-4f81-414d-9933-608cc1e58677">

**To Reproduce**
```
<mjml>
  <mj-body>
    <mj-section border="2px solid red" background-color="blue" padding="8px" border-radius="80px">
      <mj-column>
        <mj-button>
          Button
        </mj-button>
      </mj-column>
    </mj-section>
  </mj-body>
</mjml>
```

iRyusa:
Well if you inspect the HTML here you'll se that the border is not applied at the same level of the radius/background.
As we use `border-collapse: collapse;` on table to prevent issue on outlook, we can't apply it on the same element. If compatibility isn't an issue for you, you can override this behavior on the table with a `css-class` + `border-collapse: separate;` and re apply a border-radius on the `td`

arpitjp:
Thanks for the quick reply..!! Tried doing this, but didn't help. Some example I can refer?
```
<mjml>
  <mj-body>
    <mj-section border="2px solid red" background-color="blue" padding="8px" css-class="border-fix" border-radius="80px">
      <mj-column>
        <mj-button>
          Button
        </mj-button>
      </mj-column>
    </mj-section>
  </mj-body>
</mjml>
```

`<< css file >>`
```css
border-fix {
      border-collapse: separate;
    }
```

iRyusa:
Ex: https://mjml.io/try-it-live/8tUiBjdfYV

PR #2811, opened 2024-01-17
fix: border-radius is not work when i have already set border-collaps…

lzw10168:
Hello,

I have already seen this issue: https://github.com/mjmlio/mjml/issues/2741 and this ex: https://mjml.io/try-it-live/8tUiBjdfYV

But it doesn't meet my needs. In the current demo, I find myself needing to add various class names to address the border-radius for multiple Section components. What I aim for is to have the td's border-radius follow the values set on the component, eliminating the necessity for numerous class names.

I am planning to utilize [mjml](https://github.com/mjmlio/mjml) to develop a visual email style editor, allowing users to choose whether to enable border-collapse: separate !important;. I'll inform users of the associated risks while ensuring that border-radius works seamlessly on the Section component. I believe it makes logical sense to apply the border-radius solely to td, without introducing compatibility issues. If there are considerations I may have overlooked, please reject this PR. Thank you for your attention.

* Before: ` border-radius`  is not work
image

* After : Everything looks fine
image

### DIFF
--- a/packages/mjml-section/src/index.js
+++ b/packages/mjml-section/src/index.js
@@ -93,6 +93,7 @@ export default class MjSection extends BodyComponent {
         'padding-right': this.getAttribute('padding-right'),
         'padding-top': this.getAttribute('padding-top'),
         'text-align': this.getAttribute('text-align'),
+        'border-radius': this.getAttribute('border-radius'),
       },
       div: {
         ...(fullWidth ? {} : background),


iRyusa:
Thanks for the PR,
Sadly the border radius is set where it's the most "supported" without side effect on others. So you'll have to stick to css / inline css for this.

PR #2834, opened 2024-02-26
[FIX] Remove extraneous quotes around background-url

### LINKED ISSUE #2833: Extra quotes in CSS url()
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `mjml-section@4.15.3` for the project I'm working on.

I've been facing issues on Gmail with the `background-url` attribute on the `<mj-section>` component.

While it is displayed without any issue in Apple Mail (macOS) desktop, a strange behaviour happens when viewing the same email on Gmail in a web browser.

Here's a screenshot summary of the issue:

*Please ignore the flag icon issue on Gmail, that's due to Gmail not supporting SVG.*

| Description | Screenshot |
| --- | --- |
| Apple Mail, with or without this fix | Screenshot 2024-02-26 at 19 04 50 |
| Gmail, without fix. *Notice it's the same image appearing twice, though with a different crop.* | Screenshot 2024-02-26 at 19 04 35 |
| Gmail, with fix | Screenshot 2024-02-26 at 19 04 12 |

I'll open a PR later today.

In the meantime, here is the diff that solved my problem as this might help somebody encountering the same issue:

```diff
diff --git a/node_modules/mjml-section/lib/index.js b/node_modules/mjml-section/lib/index.js
index 6fca6e8..dbcb70d 100644
--- a/node_modules/mjml-section/lib/index.js
+++ b/node_modules/mjml-section/lib/index.js
@@ -91,7 +91,7 @@ let MjSection = exports.default = /*#__PURE__*/function (_BodyComponent) {
   }, {
     key: "getBackground",
     value: function getBackground() {
-      return makeBackgroundString([this.getAttribute('background-color'), ...(this.hasBackground() ? [`url('${this.getAttribute('background-url')}')`, this.getBackgroundString(), `/ ${this.getAttribute('background-size')}`, this.getAttribute('background-repeat')] : [])]);
+      return makeBackgroundString([this.getAttribute('background-color'), ...(this.hasBackground() ? [`url(${this.getAttribute('background-url')})`, this.getBackgroundString(), `/ ${this.getAttribute('background-size')}`, this.getAttribute('background-repeat')] : [])]);
     }
   }, {
     key: "getBackgroundString",
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

habovh:
Following my report on #2833 I'd like to suggest the following changes, but I believe some discussion should happen first.

As per [the CSS spec](https://developer.mozilla.org/en-US/docs/Web/CSS/url), `url()` accepts optional quotes in some cases where the URL string contains special characters. Therefore, it is reasonable for MJML to assume that we should wrap a user-provided URL in quotes, just to be on the safe side. After all, the docs says it's optional so it should work, right?

Well, Google & more specifically Gmail decided otherwise. Whether the `url()` contains a quoted value, such as `url('https://example.com/image.png')`, it will somehow affect the design.

Here's a simplified repo snippet:

```mjml
<mj-section
  background-url="{{package.cover}}"
  background-size="cover"
  background-position="center center"
  background-repeat="no-repeat"
  background-color="#d0d0d3"
>
  <mj-group>
    <mj-column width="20%">
      <mj-image
        align="left"
        src="{{package.logo}}"
      />
    </mj-column>
    <mj-column width="80%">
      <mj-image
        src="{{package.countryFlagUri}}"
        align="right"
      />
    </mj-column>
  </mj-group>
</mj-section>
```

And here's what it looks like rendered in actual email clients:

| Description | Screenshot |
| --- | --- |
| Apple Mail, with or without this fix | Screenshot 2024-02-26 at 19 04 50 |
| Gmail, without fix. *Notice it's the same image appearing twice, though with a different crop.* | Screenshot 2024-02-26 at 19 04 35 |
| Gmail, with fix | Screenshot 2024-02-26 at 19 04 12 |

So in the end I'm not 100% sure removing the quotes here is a foolproof solution, but in my case the changes from this PR allowed me to generate emails that work better on Gmail.

Important thing to know: [the MJML.io live editor](https://mjml.io/try-it-live) somehow does not add the quotes! I even installed the exact same version as the live editor (version 4.9.3 at time of writing), and locally the quotes always appeared. So I'm not sure, maybe there's some config/outdated dependency happening here? The live editor is what allowed me to diagnose the problem in the first place by diffing with my local outputs.

Lastly, if you think this change is acceptable, we might want to spread it to the `<mj-hero/>` component as well, as it seems to be using the same background-url quotes.

Any comments welcome!

Fixes #2833

### DIFF
--- a/packages/mjml-section/src/index.js
+++ b/packages/mjml-section/src/index.js
@@ -112,7 +112,7 @@ export default class MjSection extends BodyComponent {
       this.getAttribute('background-color'),
       ...(this.hasBackground()
         ? [
-            `url('${this.getAttribute('background-url')}')`,
+            `url(${this.getAttribute('background-url')})`,
             this.getBackgroundString(),
             `/ ${this.getAttribute('background-size')}`,
             this.getAttribute('background-repeat'),


iRyusa:
Took random img on google here https://mjml.io/try-it-live/BgcazlIkHS

And it looks fine on my end without the quote (4.9) :

Screenshot 2024-02-27 at 13 19 39

And with quote too (4.15) : 
Screenshot 2024-02-27 at 13 24 42

habovh:
Thanks @iRyusa for checking this out. Something weird seems to be happening here then... I'll provide a more comprehensive repro then and try to identify what's causing havoc in this specific case.

iRyusa:
Seem like quote were mandatory at some point for Yahoo/AOL https://github.com/hteumeuleu/email-bugs/issues/79 too.

I'll keep both issue+PR open for now

habovh:
Closing this PR since MJML wasn't the issue, and these changes could potentially break when using image URLs that would actually require quotes.

Mandrill/Mailchimp Transactional is the culprit here because they escape single quote characters within the inline style attributes, resulting in out-of-spec/invalid HTML. A ticket has been open with their support.

Read more about it on #2833

Issue #2856, opened 2024-04-23
Dependency: html-minifier REDoS vulnerability


boxexchanger:
**Describe the bug**
31 high severity vulnerabilities
html-minifier  *
Severity: high
kangax html-minifier REDoS vulnerability - https://github.com/advisories/GHSA-pfq8-rq6v-vf5m

**To Reproduce**
1. Init new project and install `npm i mjml`
2. check `npm audit`
result will be **31 high severity vulnerabilities**

**Expected behavior**
found 0 vulnerabilities

**Environment:**
 nodejs v20.11.1 + mjml 4.15.3

balboFK:
Confirmed. 
This needs a dependency cleanup asap. I work with mjml on most of my projects and this vulnerabilities are not good to keep.
Maybe `html-minifier` will patch the vulnerability on their end and MJML can follow up with it and patch it here.

thewilkybarkid:
`html-minifier` seems to be unmaintained. https://github.com/kangax/html-minifier/issues/1135

iRyusa:
Closing as duplicate of #2589

For what it's worth : this vulnerability isn't anything new. We started to check alternatives as html-minifier doesn't seem to be updated anymore : 
- switch to a different package html-minifier-terser is way too big as a dependency to be worth considering, htmlnano seem to be the best candidate
- forking html-minifier as `mjml-html-minifier` and patch it ourself : I think that could work too as we can just remove unused feature for MJML. I'm nowhere near an expert level in tokenizer so I don't know how to properly patch out the reDos vulnerability that `html-minifier` has. 
- removing completely minifying as an option and let users pipe the result in w/e library: not ideal IMO as some email doesn't work in some email clients when not properly minified.

You can check https://www.npmjs.com/package/mjml/v/5.0.0-alpha.1 this version remove html-minifier and js-beautify. It relies on htmlnano and prettier instead. 

```
➜  mjml-5 yarn audit
yarn audit v1.22.22
0 vulnerabilities found - Packages audited: 224
➜  mjml-5 npm audit
found 0 vulnerabilities
```

alexbjorlig:
> this vulnerability isn't anything new

Ok, great but the current version `4-15-3` explodes the terminal 💥
![CleanShot 2024-04-25 at 20 49 09@2x](https://github.com/mjmlio/mjml/assets/9967422/80388679-3ea2-44e7-a595-6a11e2e3716e)

[...]

PR #2857, opened 2024-04-24
[FIX] Replace vulnerable dependency html-minifier

marvin-wtt:
Replaces `html-minifier` in favor of `html-minifier-terser` due to a vurnability.  

Please note that I completely removed the minifer from core as it was already depricated. This however would require a new major version.


marvin-wtt:
See #2856

iRyusa:
Closing because html-minifier-terser is way too big to be considered as an option for MJML

marvin-wtt:
Maybe it's possible to publish the cli separately? The cli would than only be a dev dependency and the size wouldn't matter that much.

PR #2881, opened 2024-07-04
feat: adds direction property to mj-social-element to reverse icon/text order

jamestomasino:
image

I've been running into a lot of situations in email builds, especially on links, where designers place an arrow or other image on the right side of the text. mj-social does a fantastic job with image/text layouts, but it was lacking an option to reverse the order of the image and the text. I've added a new property `direction` as an enum of 'left' or 'right' which swaps the rendering of the TDs in the render method to support this change.

### DIFF
--- a/packages/mjml-social/README.md
+++ b/packages/mjml-social/README.md
@@ -102,6 +102,7 @@ padding-left                | px          | left offset                      | n
 padding-right               | px          | right offset                       | n/a
 padding-top                 | px          | top offset                         | n/a
 icon-padding                | px          | padding around the icon       | 0px
+icon-position               | string      | left/right                    | right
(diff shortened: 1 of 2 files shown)


iRyusa:
Hi and thanks for the contribution!

Can we rename this to `icon-position` instead ?

jamestomasino:
Absolutely. I'll modify now.

justwiebe:
When will there be a release for this? I don't see it in the latest 4.15.3 package

ColinTravis:
I'm also noticing this doesn't appear to be live in the newest `4.15.3` branch. Is there a newer version coming that will include this?

iRyusa:
Experimental branch has it and should be near a release state.

PR #2922, opened 2024-11-27
Pass parent attributes to rawXML child components

norchard:
Currently you are unable to pass attributes from a parent component to it's children components if the child is rawXML. This fix enables the parent to pass attributes to the rawXML children.


iRyusa:
Oh nice catch ! Thanks for PR 👍

PR #2948, opened 2025-04-15
added note to table docs about images and lineheight

hayden-t:


iRyusa:
I'm fine adding something like this in the doc, but it's not really related
to MJML this is how HTML/CSS and browsers works in general. mj-table isn't
the only one affected you can do the same within mj-text. Maybe it should
be mentioned in a new troubleshoot section of the doc

hayden-t:
text has a default line height of 1, not sure if that would behave similarly. but if so it might need a note too, or as you suggest have it somewhere else more 'general tips'

hayden-t:
Been a while, but this pull request suggests adding : "If you put image/s inside, you will need to set line-height at the td or table level to override the default, so it will display at desired height."

PR #2987, opened 2025-10-14
chore(testing): Update local MJML testing MJML-46

dazgreer:
Updated local testing by adding Mocha and updating existing tests to use it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "lint:fix": "eslint . --fix",
     "postinstall": "lerna bootstrap",
     "prettier": "prettier --write \"packages/**/{src,bin}/**/*.?(js|json)\"",
-    "test": "mocha packages/mjml/test/*.test.js"
+    "test": "lerna run test"
   },
   "workspaces": [
     "packages/*"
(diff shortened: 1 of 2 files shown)


guigui64 (dismissed):

package.json
```diff
@@ -8,7 +8,8 @@
     "lint": "eslint .",
     "lint:fix": "eslint . --fix",
     "postinstall": "lerna bootstrap",
-    "prettier": "prettier --write \"packages/**/{src,bin}/**/*.?(js|json)\""
+    "prettier": "prettier --write \"packages/**/{src,bin}/**/*.?(js|json)\"",
+    "test": "mocha packages/mjml/test/*.test.js"
```
guigui64: This script does exactly the same thing as the one in sub-package mjml. Is it intentional?

dazgreer: I guess not, the one at the root is the one I have been using, the one in `./packages/mjml/` was added in the last commit because of the CI build error. 

Is the latter called automatically, or did it just fail the CI build because it couldn't find the removed `./test/index.js` file? If so, I can remove the one in the package and keep the one in the root.

PR #3002, opened 2025-10-17
fix(mjml-social): align / icon-height #3001

### LINKED ISSUE #3001: mj-social attribute bugs (align and icon-height)
**Describe the bug**

1. For `mj-social-element` there is an attribute of `align` specified in the documentation however this is not set anywhere in the code when added and it doesn’t really make sense what it would do
2. When using the attribute `icon-height` on the `mj-social` tag, the `height` attribute is set both on the icon container (the background color part) and the icon itself (icon image). For most email clients, the CSS prevents the height of the icon image from taking the set value and instead this is applied to the icon container only, however in Outlook and some instances of Yahoo the image becomes stretched outside of its aspect ratio.

**To Reproduce**

1. https://mjml.io/try-it-live/dn69y69b8z
2. https://mjml.io/try-it-live/bf8oF2BHq7

**Expected behavior**
Adding the `text-align` attribute should align the text when possible, for example when `mode="vertical"` is applied
Adding `icon-height` should only affect the container (background) and not the image itself

1. https://mjml.io/try-it-live/dn69y69b8z
2. https://mjml.io/try-it-live/bf8oF2BHq7

Email on Acid / Testi

**Affected email clients (for rendering issues):**
All version of Outlook (Windows)

dazgreer:
**What:** 

1. Updated `mj-social-element` to respect `align` attribute
2. Updated `mj-social-element` so that the `icon-height` did not stretch the height of the image in Outlook

**Why:** 

1. `align` attribute was not applied
2. image height was being stretched in Outlook

**How:** 

1. attribute value is now applied
2. removed `height` attribute from the image

- Added automated test
- Tested across all supported clients in MJ 

fixes #3001

### DIFF
--- a/packages/mjml-social/src/SocialElement.js
+++ b/packages/mjml-social/src/SocialElement.js
@@ -231,7 +231,6 @@ export default class MjSocialElement extends BodyComponent {
       sizes,
       href,
       'icon-size': iconSize,
-      'icon-height': iconHeight,
     } = this.getSocialAttributes()
 
     const hasLink = !!this.getAttribute('href')


packages/mjml-social/src/SocialElement.js
```diff
@@ -263,7 +263,6 @@ export default class MjSocialElement extends BodyComponent {
                       ${this.htmlAttributes({
                         alt: this.getAttribute('alt'),
                         title: this.getAttribute('title'),
-                        height: parseInt(iconHeight || iconSize, 10),
```
totocap: We can't remove the support for icon-height, as it will be a breaking change. What you should do, maybe, is to always define height, but do not define width when icon-height is defined, to keep the aspect ratio?

dazgreer: It's not removed, I've just removed the height declaration from the image itself. The container (background color) still adheres to the height which is how it's displaying currently across almost every email client, like this: 
https://mjml.io/try-it-live/f_yv9F7oYu

This change is to prevent an issue in Outlook whereby the icon image becomes stretched, like this:
Screenshot 2025-10-17 at 10 20 51

PR #3025, opened 2025-12-15
bugfix(mjml-browser): not building

dazgreer:
**What:**
Updated `mjml-browser` package to build and work on the mjml5 branch

**Why:**
Package failed build

**How:**
- added html-minifier-terser (exc. terser) and associated dependencies
- created aliases for htmlnano and terser
- created minify.js file to process minification
- added globalObject
- updated gitignore for testing files

### DIFF
--- a/packages/mjml-browser/webpack.config.js
+++ b/packages/mjml-browser/webpack.config.js
@@ -6,6 +6,20 @@ module.exports = {
   entry: {
     "mjml": ['../mjml/lib/index'],
   },
+  externals: {
+    'cheerio': 'cheerio',
+    'undici': 'undici',
+    'cosmiconfig': 'cosmiconfig',
+  },
+  resolve: {
+    alias: {
+      'path': path.resolve(__dirname, 'browser-mocks/path'),
+      'fs': path.resolve(__dirname, 'browser-mocks/fs'),
+      'uglify-js': path.resolve(__dirname, 'browser-mocks/uglify-js'),
+      'mjml-migrate': path.resolve(__dirname, 'browser-mocks/mjml-migrate'),
+      'htmlnano': path.resolve(__dirname, 'browser-mocks/htmlnano'),
+    },
+  },
   optimization: {
     minimizer: [
       new UglifyJsPlugin({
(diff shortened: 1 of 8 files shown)


packages/mjml-browser/webpack.config.js
```diff
@@ -31,19 +60,21 @@ module.exports = {
...
+        test: /\.mjs$/,
+        include: /node_modules/,
+        type: 'javascript/auto',
+      },
       {
         test: /\.js$/,
-        exclude: path.join(__dirname, 'node_modules'),
+        exclude: [
+          path.join(__dirname, 'node_modules'),
+          /html-minifier-terser/,
```
guigui64: Why this line? Isn't `html-minifier-terser` in the `node_modules` folder and thus already excluded?

packages/mjml-browser/webpack.config.js
```diff
@@ -31,19 +60,21 @@ module.exports = {
...
-    },
+    globalObject: 'this',
   },
   module: {
     rules: [
+      {
+        test: /\.mjs$/,
+        include: /node_modules/,
+        type: 'javascript/auto',
+      },
```
guigui64: Why include `.mjs` files?

packages/mjml-browser/webpack.config.js
```diff
@@ -1,11 +1,40 @@
...
+      process: 'process/browser',
+      Buffer: ['buffer', 'Buffer'],
+    }),
+  ],
+  node: {
+    fs: 'empty',
+    global: true,
+    process: true,
+    Buffer: true,
+  },
```
guigui64, line 36: Can you explain this block?

packages/mjml-browser/webpack.config.js
DazGreerMJ: Yes it wasn't required explicitly here. 

I have updated the code though to include any node_modules except `htmlparser2` and `prettier` which still need to be processed by Babel to work here.

packages/mjml-browser/webpack.config.js
DazGreerMJ: It was a compatibility fix for Webpack 4 but it might have been on a dependency that I was trying out for minification that we're not using as removing it still allows the build. I've removed it

packages/mjml-browser/webpack.config.js
DazGreerMJ: I've updated this as well  by removing 'global' (since Webpack 4 default is true anyway) and by setting the rest to false since so it explicitly process these using the alias for 'fs' and the 'ProvidePlugin' pollyfills for the other two.

guigui64 (approved):

totocap (approved):

PR #3028, opened 2026-01-08
feature(template syntax): santization before PostCSS

dazgreer:
**What:**
Added sanitisation options for when users add templating syntax

**Why:**
PostCSS was choking on template syntax

**How:**
- Added a santizeStyles option to account for templating syntax. Can use as either complete block or CSS values either in <mj-style> tag or style=“” attributes
- Added optional templateSyntax array of objects for multiple syntax types
- Created a shim for minifyCSS to minifyCss for CLI


guigui64 (approved):

packages/mjml-core/src/index.js
```diff
@@ -41,6 +41,173 @@ import handleMjmlConfig, {
...
+    return `style="${sanitizedValue}"`
+  })
+}
+
+function restoreInlineStyleAttributes(html, syntaxes) {
+  return html.replace(/style="([^"]*)"/g, (match, styleValue) => {
+    let restoredValue = styleValue
+    syntaxes.forEach(({ prefix, suffix }, idx) => {
+      const regex = new RegExp(`sanitized${idx}:([\\s\\S]*?)(;|$)`, 'g')
+      restoredValue = restoredValue.replace(regex, `${prefix} $1 ${suffix}`)
```
totocap: Do we want to add spaces after the prefix and before the suffix? In `sanitizeInlineStyleAttributes`, maybe we could take the spaces inside the `variable` part?
Modifying `\\s*([\\s\\S]*?)\\s*` (inside the main regexp) to `(\\s*[\\s\\S]*?\\s*)`?

packages/mjml-core/src/index.js
```diff
@@ -41,6 +41,173 @@ import handleMjmlConfig, {
...
+      isBlockVariable = true
+      break
+    }
+    match = allVariablesRegex.exec(styleContent)
+  }
+
+  if (isBlockVariable && isValueVariable) {
+    throw new Error(
+      'Mixed variable syntax detected. Use either CSS property syntax (e.g., color: {{variable}}) OR block syntax (e.g., {{variable}}), not both in the same document.',
+    )
```
totocap: What do you mean? We can't do something like: `<style>{{propName}}: {{propValue}}</style>`?

packages/mjml-core/src/index.js
```diff
@@ -416,8 +585,20 @@ export default async function mjml2html(mjml, options = {}) {
...
   // PostProcessors
   if (minify) {
+    let normalizedMinifyOptions = minifyOptions
+    if (
+      minifyOptions &&
+      typeof minifyOptions.minifyCss === 'undefined' &&
+      typeof minifyOptions.minifyCSS !== 'undefined'
+    ) {
+      const legacy = minifyOptions.minifyCSS
+      const mapped = legacy === false ? false : { preset: 'lite' }
```
totocap:
```suggestion
      const mapped = minifyOptions.minifyCSS ? { preset: 'lite' } : false
```
is more readable

packages/mjml-core/src/index.js
```diff
@@ -41,6 +41,173 @@ import handleMjmlConfig, {
 
 const isNode = require('detect-node')
 
+function escapeRegex(str) {
+  return str.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')
+}
+
+function sanitizeInlineStyleAttributes(html, syntaxes) {
```
totocap, line 48: Can you add unit tests covering all possible cases, please? Even strange cases like content on multiple lines?

packages/mjml-core/src/index.js
```diff
@@ -41,6 +41,173 @@ import handleMjmlConfig, {
...
+  return html.replace(/style="([^"]*)"/g, (match, styleValue) => {
+    let sanitizedValue = styleValue
+    syntaxes.forEach(({ prefix, suffix }, idx) => {
+      const regex = new RegExp(
+        `(\\s*)${escapeRegex(prefix)}\\s*([\\s\\S]*?)\\s*${escapeRegex(suffix)}(\\s*)`,
```
Edern76, line 53: If you have something like `<div data-broken="{{var:broken" data-first="{{var:first_name}}" data-last="[[var:last_name]]">`, I think it will match `"{{var:broken" data-first="{{var:first_name}}"` as a single entity and thus replace all of this with a placeholder. 

Not sure if we have checks beforehand that prevent broken syntax like this to be passed to this function, if that's the case feel free to disregard this comment

packages/mjml-core/src/index.js
dazgreer: I've just removed the spaces when it replaces the restoredValue and this preserves them the same

packages/mjml-core/src/index.js
dazgreer: I initially had some issues when adding both block variables like <style>{{prop}}</style> alongside CSS values like <style>prop: {{propValue}}</style> but I've now resolved this and you can add both now using the optional allowMixedSyntax 

Adding properties as variables is a good shout though and you can now do that

packages/mjml-core/src/index.js
dazgreer: I've updated to this

packages/mjml-core/src/index.js
dazgreer: I've added several tests for this.

packages/mjml-core/src/index.js
dazgreer: The function only processes content either between `<style>` tags or within the `style=""` attribute so it wouldn't look at data attributes. I have added a precheck for unbalanced delimeters though even though it's not strictly necessary as the point of the sanitization is to pass PostCSS validation and this would cause it to fail and alert the user to an issue.

Edern76 (approved):

totocap (approved):

PR #3029, opened 2026-01-08
bugfix(cssnano): added lite profile to dependencies

dazgreer:
- Added the cssnoano preset as users were being prompted to install it from 5.0.0-alpha.8
- ES Lint upgrade for better compitibility with Prettier

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,9 @@
   "workspaces": [
     "packages/*"
   ],
+  "dependencies": {
+    "cssnano-preset-lite": "^4.0.4"
+  },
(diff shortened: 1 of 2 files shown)


guigui64 (approved):

package.json
```diff
@@ -14,6 +14,9 @@
   "workspaces": [
     "packages/*"
   ],
+  "dependencies": {
+    "cssnano-preset-lite": "^4.0.4"
```
totocap, line 18: What is the change in size in the bundle after passing it as a dependency?

PR #3032, opened 2026-01-14
bugfix(ignoreIncludes): security fix #3018

### LINKED ISSUE #3018: [CVE-2025-67898] CVE-2020-12827 unfixed as `ignoreIncludes` continues to default to `false`
Hi!

Past vulnerability CVE-2020-12827 has come to my attention. After reading [the detailed description](https://seclists.org/fulldisclosure/2020/Jun/23) and inspecting the two related commits…

- 21034a7187f2bda4b0d66e7a1c63314c88dce71a
- 30e29ed2cdaec8684d60a6d12ea07b611c765a12

…[between releases 4.6.2 and 4.6.3](https://github.com/mjmlio/mjml/compare/v4.6.2...v4.6.3), my impression so far is that the issue of attacker-controlled file access has not been fixed and that the default settings are still vulnerable. 

For proof (1), in version `<mj-include path="/etc/passwd" />` renders an error while `<mj-include path="no_such_file" />` renders a comment `<!-- mj-include fails to read file [..] -->` and overall rendering succeeds, so checking for the existence of files still works (provided some error feedback channel to the attacker).

What's worse (2), when adding `type="css"` as documented at https://documentation.mjml.io/#mj-include a la …

```xml
<mjml>
  <mj-body>
    <mj-include path="/etc/passwd" type="css" />
  </mj-body>
</mjml>
```

…and compiling via `mjml -r demo.mjml` the whole file content is included verbatim into the HTML header, and no error is produced.

So from my point of view CVE-2020-12827 is wide open by default and has never been fixed. Could you clarify and point me to what I am missing?

Thanks and best, Sebastian

CC @hannob

dazgreer:
- set ignoreIncludes to default to true
- setflag for CLI to set to true
- updated docs

Fixes #3018

### DIFF
--- a/packages/mjml-core/src/index.js
+++ b/packages/mjml-core/src/index.js
@@ -111,7 +111,7 @@ export default function mjml2html(mjml, options = {}) {
     keepComments,
     minify = false,
     minifyOptions = {},
-    ignoreIncludes = false,
+    ignoreIncludes = true,
     juiceOptions = {},
     juicePreserveTags = null,
     skeleton = defaultSkeleton,
--- a/packages/mjml-cli/README.md
+++ b/packages/mjml-cli/README.md
@@ -155,6 +155,16 @@ $> mjml ./my-project/input.mjml --config.filePath ./my-partials/
 
 If you like to keep your partials together and you want to be able to mj-include them without having to change the relative path of the includes depending on the compiled file path, you can use this option. In this exemple, `<mj-include path="./header.mjml" />` will include `./my-partials/header.mjml`, ignoring the actual path of `input.mjml`.
 
+### Allow includes (opt-in)
+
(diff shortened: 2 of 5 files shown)


packages/mjml-core/src/index.js
```diff
@@ -111,7 +111,7 @@ export default function mjml2html(mjml, options = {}) {
     keepComments,
     minify = false,
     minifyOptions = {},
-    ignoreIncludes = false,
+    ignoreIncludes = true,
```
totocap, line 114: This is a breaking change, it should be added into MJML 5, no? An explain to our users

dazgreer:
Closing as we'll merge these changes in MJML 5.

New PR: https://github.com/mjmlio/mjml/pull/3033

PR #3038, opened 2026-01-19
chore(lerna): move to lerna v9

totocap:
Reverts mjmlio/mjml#3037


totocap:
Could no run `yarn lerna version` with success. No error logged.

PR #3044, opened 2026-01-29
bugfix(attributes): adding/updating various for consistency

dazgreer:
Added or updated attributes across several packages that were missing or inconsistent

- gave border-radius consistent input across all packages
- updated inner-padding attributes of mj-hero and fixed an Outlook issue with width/padding
- updated all docs to reflect changes


packages/mjml-button/src/index.js
```diff
@@ -12,7 +12,7 @@ export default class MjButton extends BodyComponent {
     'background-color': 'color',
     'border-bottom': 'string',
     'border-left': 'string',
-    'border-radius': 'string',
+    'border-radius': 'unit(px,%){1,4}',
```
totocap: This is a breaking change, it needs to be listed with the other breaking changes for MJML 5. What about a migration guide, to help people migrate from MJML 4 to MJML 5?

packages/mjml-hero/src/index.js
```diff
@@ -15,8 +15,7 @@ export default class MjHero extends BodyComponent {
     'background-width': 'unit(px,%)',
     'background-height': 'unit(px,%)',
     'background-position': 'string',
-    'border-radius': 'string',
-    'container-background-color': 'color',
```
totocap, line 19: `container-background-color` was deleted because not used?

totocap (approved):

packages/mjml-hero/src/index.js
DazGreerMJ: this attribute doesn't appear to be used by the code at all. there is a reference to `container-background-color` on lines 213 and 218 but these are read from child components. This may have been a change at some point but the the attribute not removed.

packages/mjml-button/src/index.js
DazGreerMJ: I'm having a rethink about this. Currently there is a mix of values allowed for `border-radius` across components, either `string` or `unit(px,%)` or `unit(px,%){1,4}`. I wanted to harmonise these for consistency but actually the stricter value does not allow, e.g. `100px 50px 100px 50px / 50px 50px 50px 50px` which is valid in CSS and gives an asymmetrical border radius. I'm considering whether they should all be reverted to 'string' for simplicity.

dazgreer: I've updated to allow string across all instances of border radius. This makes it consistent across packages, allows valid input that would previously errored and removes the breaking change

totocap (approved):

PR #3045, opened 2026-01-29
feature(ignoreIncludes): added option for includePath

### LINKED ISSUE #3018: [CVE-2025-67898] CVE-2020-12827 unfixed as `ignoreIncludes` continues to default to `false`

dazgreer:
- This option allows you to specify include folders outside of your filePath whilst maintaining the sandbox security
- Added fullyDecode() loop to decode double/triple-encoded include paths
- Early reject includes with null bytes, absolute paths, UNC paths, or Windows drive letters.
- Move decoding into handleInclude() and handleCssHtmlInclude() . avoid event-time decode; support only mjml, html and css
- Added testing
- Updated docs

Fixes #3018


packages/mjml-core/src/index.js
```diff
@@ -443,13 +444,53 @@ export default async function mjml2html(mjml, options = {}) {
...
+      }
+      const baseReal = fs.existsSync(base) ? fs.realpathSync(base) : base
+      for (const p of pathsArr) {
+        if (fs.existsSync(p)) {
+          const r = fs.realpathSync(p)
+          const relToBase = path.relative(baseReal, r)
+          const isOutsideBase =
+            !relToBase ||
+            relToBase.startsWith('..') ||
+            path.isAbsolute(relToBase)
```
totocap, line 476: Does it cover cases like `./templates/../../headers`? So moving outside after entering in a folder.

packages/mjml/test/include-path-cli-cwd.test.js
```diff
@@ -0,0 +1,7 @@
+// Removed: CWD-based sandbox test. Recommended approaches:
+// - Set --config.filePath to project templates root
+// - Or use --config.includePath to explicitly allow sibling directories
+// This test was removed to avoid ambiguity in relative include resolution.
+describe('CLI CWD sandbox allows sibling includes without includePath', function () {
+  it('skipped - see include-path-cli.test.js for supported CLI usage', function () {})
+})
```
totocap: What is the purpose of this file?

packages/mjml/test/include-path.test.js
```diff
@@ -0,0 +1,131 @@
...
+      header: path.relative(rootDir, path.join(common1Dir, 'header.mjml')),
+      footer: path.relative(rootDir, path.join(common2Dir, 'footer.mjml')),
+      css: path.relative(rootDir, path.join(common1Dir, 'styles.css')),
+      html: path.relative(rootDir, path.join(common2Dir, 'snippet.html')),
+    }
+
+    const template = buildTemplate(paths)
+
+    const { html } = await mjml(template, {
+      filePath: rootDir,
```
totocap, line 115: Should we had tests for includePath, too?

dazgreer: that test is above this one (line 86)? Or did you mean something else?

packages/mjml/test/include-path-cli-cwd.test.js
dazgreer: Not required and removed.

packages/mjml-core/src/index.js
dazgreer: includes are resolved with `fs.realpathSync` and checked via `path.relative` against allowed roots. Any target whose resolved path is outside `filePath` or the configured `includePath` entries is denied, so traversal is blocked. I've added a test for this

totocap (approved):

totocap (approved):

PR #3055, opened 2026-02-28
doc: add mjml2amp to ports (AMP for Email)

m-Ryan:
- Live demo https://www.easyemail.pro/

### DIFF
--- a/doc/ports.md
+++ b/doc/ports.md
@@ -32,6 +32,12 @@ Ruby wrapper for MRML, the MJML markup language implementation in Rust.
 
 [https://github.com/hardpixel/mrml-ruby](https://github.com/hardpixel/mrml-ruby)
 
+### AMP for Email: mjml2amp
+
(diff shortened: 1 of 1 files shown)


DazGreerMJ (commented):
Hi, thanks for the PR for this potentially useful tool.

I'd like to hold off adding this to our documentation for a few reasons:

1. It's based on 5.0.0-alpha.9. MJML 5 subsequently evolved and received important dependency and security updates. It would be great if you could validate it against the latest 5.4.0 and confirm whether the intention is to track the current MJML version.
2. The latest 1.1.0 release has a CI failure. The logs have expired, so I'm not sure what caused it.
3. The README published to npm is the old MJML 4 README.
4. The wording states AMP for Email compliant. I can't find any evidence of automated AMP validation/testing, so perhaps compatible would be more appropriate than compliant?

Once these points are clarified, I'd be happy to take another look. Thanks

PR #3065, opened 2026-03-31
Implement path traversal protection in MJML includes

zdods:
- Added security checks to prevent relative and absolute path traversal in `mj-include` tags, ensuring included files are within the project directory.
- Introduced a new helper function `assertPathWithinRoot` to validate file paths.
- Updated documentation to reflect the new security measures.
- Added tests to verify the enforcement of these security rules and ensure error messages do not leak filesystem paths.


zdods:
Fixed in 5.0.0

PR #3075, opened 2026-04-22
Bugfix/beautification using prettier is long

totocap:
- The call to beautify took 500ms for big MJML document since we were using
  Prettier instead of 15ms with js-beautify. We can't go back to js-beautify
because of security issues so we are now using Biome (with its WASM Node.js
port) to beautify the HTML file. Biome creates a file in memory, not on the
disk.
- Still uses prettier for the beautify option of mjml-browser.
- Add beautify unit tests.


packages/mjml-core/src/index.js
```diff
@@ -873,10 +875,18 @@ export default async function mjml2html(mjml, options = {}) {
...
-      parser: 'html',
-      printWidth: 240,
-    })
+    if (isNode) {
+      const formatHtml =
+        nodeFormatter.formatHtml ||
+        (nodeFormatter.default && nodeFormatter.default.formatHtml)
+      content = formatHtml(content)
+    } else {
+      // Browser fallback keeps prettier formatting behavior.
```
guigui64: Why not use biome WASM in the browser as well? There's a `@biomejs/wasm-web` package.

packages/mjml-core/src/index.js
```diff
@@ -12,9 +12,8 @@ import {
 } from 'lodash'
 import juice from 'juice'
 import { load } from 'cheerio'
-import prettier from 'prettier'
 import minifier from 'htmlnano'
-
+import prettier from 'prettier'
```
dazgreer: Can this (or wasm-web if we switch to that) be lazy loaded inside mjml-browser since it's only used there?

packages/mjml-core/src/index.js
```diff
@@ -873,10 +875,18 @@ export default async function mjml2html(mjml, options = {}) {
       }
     }
   } else if (beautify) {
```
dazgreer, line 964: We discussed that Biome doesn’t do parsing for language specific syntax but we can add our template syntax sanitization inside the `beautify` block here (as it is inside the `minify` block

packages/mjml-core/src/index.js
```diff
@@ -873,10 +875,18 @@ export default async function mjml2html(mjml, options = {}) {
...
-    content = await prettier.format(content, {
-      parser: 'html',
-      printWidth: 240,
-    })
+    if (isNode) {
+      const formatHtml =
+        nodeFormatter.formatHtml ||
+        (nodeFormatter.default && nodeFormatter.default.formatHtml)
+      content = formatHtml(content)
```
dazgreer, line 991: Should this be called with `await`?

```
formatHtml in node-formatter.js doesn't return a Promise so 
this works today, but the function signature in the browser 
mock is async function formatHtml. The two are inconsistent.
The call site should be content = await formatHtml(content) 
and node-formatter.js should export an async function - 
both for consistency with the browser mock and to future-proof 
if the WASM API becomes async.
```

packages/mjml-core/src/index.js
totocap: I tried with @biomejs/wasm-web, several things, but it didn't work. Our Webpack version is a bit old (we will need to upgrade to 5) and the bundle will need to include a .wasm file that will need to be loaded by the applications that uses mjml-browser. I don't think it is a good idea. We can replace Prettier by something else if we want, by it needs to be JS only. Also, WASM files are huge, generally

packages/mjml-core/src/index.js
totocap: Done

packages/mjml-core/src/index.js
totocap: Done

packages/mjml-core/src/index.js
totocap: Removed the async from the mock.

guigui64 (approved):

guigui64 (approved):

PR #3076, opened 2026-04-23
bugfix(core): additional beautify support

dazgreer:
Biome formatter
- Split off any content before `<!doctype html>` (e.g. `mj-raw position="file-start" `template tags) before passing to Biome, then prepend it back. Biome's WASM parser panics on documents that don't start with a doctype.
- Enabled `formatWithErrors`: true so recoverable parse errors don't abort formatting of the rest of the document.
- Updated Biome options and added CSS formatter config

Beautify path improvements
- Template-syntax sanitization (`sanitizeStyles`) now also applies when `beautify: true`, mirroring the existing `minify: true` behaviour.
- 
htmlmin:ignore re-implemented
- html-minifier was replaced by htmlnano in a previous PR; Re-implemented support by extracting each `<!-- htmlmin:ignore -->…<!-- htmlmin:ignore -->` pair into an opaque token before htmlnano runs, then restoring it afterwards.
- Stripped `<!-- htmlmin:ignore -->` markers from beautified output

Browser bundle
- Extend webpack babel-loader rule from `/\.js$/` to `/\.m?js$/` so that prettier/plugins/html (an ESM .mjs file) is transpiled correctly.
- Require `prettier/plugins/html` explicitly and pass it in the plugins array

CLI
- `allowMixedSyntax` now gets the same string→boolean normalisation

Updates to docs and automated testing

### DIFF
--- a/packages/mjml-core/src/index.js
+++ b/packages/mjml-core/src/index.js
@@ -918,6 +918,22 @@ export default async function mjml2html(mjml, options = {}) {
       ]
 
     const cssMinifyEnabled = htmlnanoOptions.minifyCss !== false
+
+    // Protect content wrapped in <!-- htmlmin:ignore --> pairs from whitespace
+    // collapsing. We extract each protected chunk into an opaque token so
+    // htmlnano never sees the content, then restore it afterwards.
+    const htmlminIgnoreMap = []
+    if (content.includes('htmlmin:ignore')) {
+      content = content.replace(
+        /<!--\s*htmlmin:ignore\s*-->([\s\S]*?)<!--\s*htmlmin:ignore\s*-->/g,
+        (_, inner) => {
+          const token = `MJMLHTMLIGNORE${htmlminIgnoreMap.length}END`
+          htmlminIgnoreMap.push(inner)
+          return token
+        },
+      )
+    }
+
     const sanitizationResult = sanitizeTemplateVariablesInHtml(
       content,
       sanitizeStyles === true && cssMinifyEnabled,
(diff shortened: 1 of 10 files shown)


packages/mjml-core/src/index.js
```diff
@@ -918,6 +918,22 @@ export default async function mjml2html(mjml, options = {}) {
       ]
 
     const cssMinifyEnabled = htmlnanoOptions.minifyCss !== false
+
+    // Protect content wrapped in <!-- htmlmin:ignore --> pairs from whitespace
+    // collapsing. We extract each protected chunk into an opaque token so
+    // htmlnano never sees the content, then restore it afterwards.
+    const htmlminIgnoreMap = []
+    if (content.includes('htmlmin:ignore')) {
```
totocap: No need to check if the content includes `htmlmin:ignore`, the regexp will do nothing if it is not included. Otherwise you will go over the HTML two times.

packages/mjml-core/src/index.js
```diff
@@ -918,6 +918,22 @@ export default async function mjml2html(mjml, options = {}) {
       ]
 
     const cssMinifyEnabled = htmlnanoOptions.minifyCss !== false
+
+    // Protect content wrapped in <!-- htmlmin:ignore --> pairs from whitespace
+    // collapsing. We extract each protected chunk into an opaque token so
+    // htmlnano never sees the content, then restore it afterwards.
+    const htmlminIgnoreMap = []
```
totocap: That's not a map, that's an array, so `htmlminIgnoreList`?

totocap (approved):

PR #3079, opened 2026-04-30
feature(beautify): Biome to optional dep + other beautify updates

### LINKED ISSUE #3077: v5.1.0 increased node_modules dependency size by 63%
We were surprised when upgrading to 5.1.0 that our node_modules size increased by about 30 megs. It looks like the @biomejs package added the following:
```
33208	node_modules/@biomejs
```

This increased the total install size from 52,340 to 85,560 (as reported by du -k).

We currently use mjml for email generation in a back-end service via AWS lambda and this increase pushed us over some limits which is how we noticed it.

Not reporting this as a bug or a feature request but more of an informational issue since if you intend to stay with @biomejs then we will have to do something different.

### LINKED ISSUE #3078: Try it Live seems to add spaces to mj-text block causing improper formatting for anchor tags when immediately followed by punctuation
**Describe the bug**
I usually use the Try it Live tool to compile my mjml ->html. Yesterday I noticed that if an anchor tag is followed by punctuation there is a space between the link and the punctuation. When I use the Export HTML tool in vscode the spaces are not present and the html compiles and renders as expected.

**To Reproduce**
```
<mjml>
  <mj-head></mj-head>
  <mj-body>
    <mj-section>
      <mj-column>
        
        <mj-text>
          Styling has been removed from this block to help root out possible
          causes of spacing issue after
          <a href="https://www.mjml.io">anchor tag</a>.
        </mj-text>

        <mj-text
          font-family="'Open Sans', 'Segoe UI', sans-serif"
          mj-class="title"
          font-size="16px"
          line-height="24px"
        >
          For more information about the our email,
          <a href="https://www.mjml.io" class="link-inline"
            >visit their website</a
          >, or contact Luke Skywalker at
          <a href="tel:+15550001234" class="link-inline">his phone number</a>,
          or <a href="mailto:bigdog@mjml.io" class="link-inline">my email</a>.
        </mj-text>
      </mj-column>
    </mj-section>
  </mj-body>
</mjml>
```
2. Render it to HTML by adding it to the Try it Live tool.
3. See error

Or create a file with the same code and export html (not using Try it Live) to see the difference.

**Expected behavior**
I expect there to be no space between a link and punctuation (if punctuation immediately follows a link).

 - OS: Windows 11 Pro, v23H2, os build 22631.6199
 - MJML Version: VSCode extension v2.2.2
 - MJML tool used: VSCode  -> Try it Live v5.1.0

 - Platform used to send the email: [e.g [Constant Contact](https://constantcontact.com/)], but after digging into it the bug seems to originate at Try it Live 

**Screenshots**
How test email should look
Image
How test email looks in vscode
Image
Improper rendering in Try it Live
Image
Improper rendering of Try it Live html in Constant Contact
Image
Proper rendering using Export HTML from VSCode using MJML extension
Image

**Additional context**
Using Export HTML from VSCode is a good workaround for now, but curious what's going on.

dazgreer:
### Features
- Biome now an optional dependency so users can choose to exclude it. Added graceful error  (Fixes: #3077)
- `beautify` option now allows a function for users to add their own beautification process. Similar in CLI using `beautifyModule` argument

### Fixes
- Changed `whitespaceSensitivity` setting to `css` to prevent issue with Biome adding a space between inline tags and punctuation. Added some normalisation to fix some issues with whitespace in <a> and <label> tags due to the change (Fixes: #3078)

### Chores
- Moved Prettier out of `mjml-core` dependencies and into `mjml-browser`

### Testing
- Added automated testing for Beautify as a custom function

### Testing
- Updated docs to add new `beautify` options


guigui64 (approved):

terencehonles:
Can we have an option to disable biome and use prettier? I don't think it's ready for prime time (or at least this project). Prettier seems to work OK.

Example input:
```{% for item in items %}{% endfor %}```

Biome output:
```{ % for item in items %} { % endfor %}```

Prettier output:
```{% for item in items %}{% endfor %}```

[Biome playground link](https://biomejs.dev/playground/?tab=formatter&pane=Diagnostics&code=ewAlACAAZgBvAHIAIABpAHQAZQBtACAAaQBuACAAaQB0AGUAbQBzACAAJQB9AHsAJQAgAGUAbgBkAGYAbwByACAAJQB9AAoA&language=html)

The spacing being added between the "{" and "%" is significant because we are running a post processing template step (using `config.templateSyntaxes` does not seem to help), and the extra whitespace in the for loop has meaning (for this trivial example it makes no sense to write the code given, but adding whitespace in a loop or shifting around markers will mean something and should not be done.

Here is a longer example that shows the code above is not a problem because it's not in an HTML > BODY tag:

```html
<!doctype html>
<html
	lang="und"
	dir="auto"
	xmlns="http://www.w3.org/1999/xhtml"
	xmlns:v="urn:schemas-microsoft-com:vml"
	xmlns:o="urn:schemas-microsoft-com:office:office"
>
	<head>
		<title></title>
		<!--[if !mso]><!-->
		<meta http-equiv="X-UA-Compatible" content="IE=edge" />
		<!--<![endif]-->
		<meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
		<meta name="viewport" content="width=device-width, initial-scale=1" />
		<style type="text/css"></style>
	</head>
	<body style="word-spacing: normal; background-color: #f8f9fa">
		<div
			aria-roledescription="email"
			role="article"
			lang="und"
			dir="auto"
			style="word-spacing: normal; background-color: #f8f9fa"
		>
			{% for item in items %}{% endfor %}
		</div>
		<span>{% for item in items %}{% endfor %}</span>
	</body>
</html>
```

[Biome playground link](https://biomejs.dev/playground/?tab=formatter&pane=Diagnostics&code=PAAhAGQAbwBjAHQAeQBwAGUAIABoAHQAbQBsAD4ACgA8AGgAdABtAGwACgAJAGwAYQBuAGcAPQAiAHUAbgBkACIACgAJAGQAaQByAD0AIgBhAHUAdABvACIACgAJAHgAbQBsAG4AcwA9ACIAaAB0AHQAcAA6AC8ALwB3AHcAdwAuAHcAMwAuAG8AcgBnAC8AMQA5ADkAOQAvAHgAaAB0AG0AbAAiAAoACQB4AG0AbABuAHMAOgB2AD0AIgB1AHIAbgA6AHMAYwBoAGUAbQBhAHMALQBtAGkAYwByAG8AcwBvAGYAdAAtAGMAbwBtADoAdgBtAGwAIgAKAAkAeABtAGwAbgBzADoAbwA9ACIAdQByAG4AOgBzAGMAaABlAG0AYQBzAC0AbQBpAGMAcgBvAHMAbwBmAHQALQBjAG8AbQA6AG8AZgBmAGkAYwBlADoAbwBmAGYAaQBjAGUAIgAKAD4ACgAJADwAaABlAGEAZAA%2BAAoACQAJADwAdABpAHQAbABlAD4APAAvAHQAaQB0AGwAZQA%2BAAoACQAJADwAIQAtAC0AWwBpAGYAIAAhAG0AcwBvAF0APgA8ACEALQAtAD4ACgAJAAkAPABtAGUAdABhACAAaAB0AHQAcAAtAGUAcQB1AGkAdgA9ACIAWAAtAFUAQQAtAEMAbwBtAHAAYQB0AGkAYgBsAGUAIgAgAGMAbwBuAHQAZQBuAHQAPQAiAEkARQA9AGUAZABnAGUAIgAgAC8APgAKAAkACQA8ACEALQAtADwAIQBbAGUAbgBkAGkAZgBdAC0ALQA%2BAAoACQAJADwAbQBlAHQAYQAgAGgAdAB0AHAALQBlAHEAdQBpAHYAPQAiAEMAbwBuAHQAZQBuAHQALQBUAHkAcABlACIAIABjAG8AbgB0AGUAbgB0AD0AIgB0AGUAeAB0AC8AaAB0AG0AbAA7ACAAYwBoAGEAcgBzAGUAdAA9AFUAVABGAC0AOAAiACAALwA%2BAAoACQAJADwAbQBlAHQAYQAgAG4AYQBtAGUAPQAiAHYAaQBlAHcAcABvAHIAdAAiACAAYwBvAG4AdABlAG4AdAA9ACIAdwBpAGQAdABoAD0AZABlAHYAaQBjAGUALQB3AGkAZAB0AGgALAAgAGkAbgBpAHQAaQBhAGwALQBzAGMAYQBsAGUAPQAxACIAIAAvAD4ACgAJAAkAPABzAHQAeQBsAGUAIAB0AHkAcABlAD0AIgB0AGUAeAB0AC8AYwBzAHMAIgA%2BADwALwBzAHQAeQBsAGUAPgAKAAkAPAAvAGgAZQBhAGQAPgAKAAkAPABiAG8AZAB5ACAAcwB0AHkAbABlAD0AIgB3AG8AcgBkAC0AcwBwAGEAYwBpAG4AZwA6ACAAbgBvAHIAbQBhAGwAOwAgAGIAYQBjAGsAZwByAG8AdQBuAGQALQBjAG8AbABvAHIAOgAgACMAZgA4AGYAOQBmAGEAIgA%2BAAoACQAJADwAZABpAHYACgAJAAkACQBhAHIAaQBhAC0AcgBvAGwAZQBkAGUAcwBjAHIAaQBwAHQAaQBvAG4APQAiAGUAbQBhAGkAbAAiAAoACQAJAAkAcgBvAGwAZQA9ACIAYQByAHQAaQBjAGwAZQAiAAoACQAJAAkAbABhAG4AZwA9ACIAdQBuAGQAIgAKAAkACQAJAGQAaQByAD0AIgBhAHUAdABvACIACgAJAAkACQBzAHQAeQBsAGUAPQAiAHcAbwByAGQALQBzAHAAYQBjAGkAbgBnADoAIABuAG8AcgBtAGEAbAA7ACAAYgBhAGMAawBnAHIAbwB1AG4AZAAtAGMAbwBsAG8AcgA6ACAAIwBmADgAZgA5AGYAYQAiAAoACQAJAD4ACgAJAAkACQB7ACUAIABmAG8AcgAgAGkAdABlAG0AIABpAG4AIABpAHQAZQBtAHMAIAAlAH0AewAlACAAZQBuAGQAZgBvAHIAIAAlAH0ACgAJAAkAPAAvAGQAaQB2AD4ACgAJAAkAPABzAHAAYQBuAD4AewAlACAAZgBvAHIAIABpAHQAZQBtACAAaQBuACAAaQB0AGUAbQBzACAAJQB9AHsAJQAgAGUAbgBkAGYAbwByACAAJQB9ADwALwBzAHAAYQBuAD4ACgAJADwALwBiAG8AZAB5AD4ACgA8AC8AaAB0AG0AbAA%2BAAoA&language=html)

dazgreer:
Hi @terencehonles, 

Prettier has been removed completely from the Node/CLI installation and unfortunately we can't use it due to performance issues. 

You can disable `beautify` completely in the current 5.1.0 version. Unless you specifically need to use the HTML output then there is no need to beautify at all.

This PR also introduces the possibility to exclude the Biome install and additionally to add your own beautification process, which may work for your needs. If Prettier works for you then you can still use that, you'll just need to pass it as a function to the `beautify` option.

The other option I'm considering is whether a `<!-- htmlbeautify:ignore -->` comment would be useful, which would work similarly to how the existing `<!-- htmlmin:ignore -->` tag does for minification and ignore anything between it. So you would setup your code like this:

```
<mjml>
  <mj-body>
    <mj-section>
      <mj-column>
        <!-- htmlbeautify:ignore -->{% for item in items %}{% endfor %}<!-- htmlbeautify:ignore -->
      </mj-column>
    </mj-section>
  </mj-body>
</mjml>
```

terencehonles:
If that's the case it seems like running "beautify" should be disabled by default...

I'm surprised biome mangles the HTML and I'm not sure why (even if it's text it **does** mean something and whitespace is being added in places it shouldn't). I don't really want to invest the time in figuring out why biome isn't working, as I'm only using it because mjml is -- I did like the idea of reformatting the output since we have it checked in and I figured it might make it easier to diff.

I have already disabled the beautification process, but I'd rather not have had to figure out how to do that and figured I'd report that there were other errors with biome than you may have been aware of.

dazgreer:
Hi @terencehonles, we appreciate the feedback!

`beautify` is disabled by default when running through Node (but not in CLI or some other areas of our ecosystem) so it depends how you consume it. This may change longer term but it would be a breaking change. 

Biome is outputting HTML the same as our previous beautifiers but templating languages are not HTML which seems to be the issue and Biome is not handling it as expected in some instances. Disabling beautify is the best solution if you are using template syntax inside your HTML but we are considering whether introducing comments to allow you to still beautify whilst skipping template syntax would be a useful tool too.

terencehonles:
Using biome itself seems to be a breaking change and that was done in this most recent **minor** release. So that comment doesn't seem to be consistent with what happened.

We do indeed use the CLI, so that would apply to us and why biome is being used.

I understand the example I'm giving is a templating language, but if I were not post-processing this and trying to show an example of some code or even just the string `{%` it would render as `{ %` which is not correct. A `<span>{%</span>` should never have any new white space because that is an inline tag and would not have the spaces collapsed (and this collapsing only applies to multiple whitespace characters, not one without any spaces in it). I'm not sure what other text biome is mangling, but I don't think this project should be using it by default anywhere as it is not required and may change the output in a way that is unexpected and hard to notice.

Edit: it looks like biome adds a space between `{` and any text as long as it's not `{{`.

dazgreer:
After consideration, we agree and subsequently we've released v5.2.0 which resolves this issue by moving back to js-beautify (our original beautifier) and output, both HTML and template syntax should now be consistent with v4. 

There's more detail on this in the release notes: https://github.com/mjmlio/mjml/releases/tag/v5.2.0

PR #3081, opened 2026-05-04
feature(beautify): added comments to exclude content from process

dazgreer:
- Added  comment to wrap around content and exclude it from the beautify process


dazgreer:
Closing as we moved in a different direction: https://github.com/mjmlio/mjml/releases/tag/v5.2.0

PR #3082, opened 2026-05-05
refactor(core,browser): switch beautification to js-beautify

### LINKED ISSUE #3077: v5.1.0 increased node_modules dependency size by 63%

### LINKED ISSUE #3078: Try it Live seems to add spaces to mj-text block causing improper formatting for anchor tags when immediately followed by punctuation

dazgreer:
## What's changed
- Switched beautification implementation to js-beautify in core formatting flow.  (Fixes: #3077, Fixes: #3078)
- Simplified Node-only formatter module and preserved file-start raw safety handling.
- Updated browser webpack handling for js-beautify.
- Removed obsolete mjml-browser artifacts (uglify-js alias/mock, path mock)
- Updated tests and docs

## Testing
- HTML rendering is identical to v4.18.0 which used the same beautifier
- Template tokens are kept intact
- Automated testing and browser build all pass


jbsinch (approved):

packages/mjml-core/src/index.js
```diff
@@ -983,22 +983,23 @@ export default async function mjml2html(mjml, options = {}) {
...
+      const { html: beautify } = require('js-beautify')
+      content = beautify(content, {
+        indent_size: 2,
+        indent_char: ' ',
+        max_preserve_newlines: 1,
+        preserve_newlines: true,
+        unformatted: [],
+        wrap_line_length: 0,
+        wrap_attributes: 'auto',
       })
```
totocap: Do we need to different paths now? Also, with this implementation, we will not apply the "doctype fix" we have inside `formatHtml`

packages/mjml/test/beautify-output.test.js
```diff
@@ -284,7 +286,7 @@ describe('Beautify output', function () {
       chai.expect(plainHtml.slice(0, plainDoctypeIndex)).to.include(line)
       chai.expect(beautifiedHtml.slice(0, beautifiedDoctypeIndex)).to.include(line)
     })
-    chai.expect(beautifiedHtml).to.include('\n  <head>\n')
+    chai.expect(beautifiedHtml).to.match(/\n\s*<head>\s*\n/)
```
totocap, line 289: Can we add a test with the Twig example?

packages/mjml-core/src/node-only/node-formatter.js
```diff
@@ -1,94 +1,31 @@
...
-      },
-    },
-    workspace_directory: '/',
-  })
-}
+const { html: beautify } = require('js-beautify')
 
 /**
- * Format HTML using Biome WASM (Node.js only).
+ * Format HTML using js-beautify (Node.js only).
```
totocap: Not "(Node.js only)" anymore?

DazGreerMJ: This file is removed as per the above refactor

packages/mjml/test/beautify-output.test.js
DazGreerMJ: Added a test to beautify-output.test.js

packages/mjml-core/src/index.js
DazGreerMJ: I've extracted js-beautify HTML formatting into a shared helper and removed the node-only/node-formatter.js file

totocap (approved):

PR #3092, opened 2026-05-18
fix(mjml-cli): resolve relative includePath, warn on denied mj-includes

### LINKED ISSUE #3090: Unable to use config.includepath to get <mj-include> to work
Having a very frustrating time with upgrading to 5.X getting the new includes security to work, I managed to get it to work in the VSCode plugin but I am unable to get the CLI tool to work. Here is my test templates setup:

```
layouts/header.mjml
layouts/styles.mjml
templates/template.mjml
```

in template.mjml my includes are `<mj-include path="../layouts/styles.mjml" />` and `<mj-include path="../layouts/header.mjml" />`

when I run
`mjml templates/template.mjml` from the root the includes do not work so I updated it to use `mjml templates/template.mjml --config.allowIncludes true --config.includePath '["layouts"]'` but in the output I get `<!-- mj-include denied -->`
I tried
`mjml templates/template.mjml --config.allowIncludes true --config.includePath '["/layouts"]'`
as well
But nothing is working and the docs are not helpful

### LINKED ISSUE #3091: Improve examples documentation for MJML 5 defaults, or a warning?
I was looking at the [components_2.md example](https://github.com/mjmlio/mjml/blob/master/doc/components_2.md) on how to create MJML components with the `<mj-include>` directive and it was failing silently on my `mjml` cli invocation. I wasn't seeing the header even though I used `mj-include` to bring it in. After some Googling and prompting, I then saw that MJML 5 disables includes by default. Once I added `allowIncludes: true` to the config, it worked. Could the documentation be enhanced so that any changes in MJML 5 are reflected?

In addition, I thought it might be a good idea to have a warning message in the CLI when it sees an `mj-include` and you haven't enabled them. Maybe a warning that says something like "mj-include is disabled by default." with some kind of helpful suggestion on either enabling them, or using a different pattern altogether.

dazgreer:
## What's changed?

Fixes two issues with mj-include path handling in the CLI 

- Relative `includePath` entries are now resolved against `process.cwd()` before being passed to mjml-core, so `--config.includePath '["layouts"]'` works correctly when the template lives in a subdirectory. (Fixes: #3090)

- The false-positive `[MJML security]` console warning that fired for every valid `includePath` entry outside the template directory has been removed. The warning is now scoped to the root filesystem path only.

- Two new CLI warnings are emitted to stderr:
  - When includes are disabled (default) but `mj-include` tags are present in the input, users are told to add `--config.allowIncludes true`.
  - When includes are enabled but one or more includes were denied (outside allowed roots), users are told to add the missing directory to `--config.includePath`. (Fixes: #3091)

- Regression tests added for all new behaviours.

- Documentation updated


packages/mjml-cli/src/client.js
```diff
@@ -154,6 +154,22 @@ export default async () => {
     }
   }
 
+  // Resolve relative includePath entries to absolute paths using the CLI's
+  // working directory (process.cwd()), not the template file's directory.
+  // This matches the intuitive expectation that `--config.includePath '["layouts"]'`
+  // refers to a folder relative to where the CLI command is run.
+  const resolveIncludePathEntry = (p) => {
+    if (typeof p === 'string' && p.length > 0 && !path.isAbsolute(p)) {
+      return path.resolve(process.cwd(), p)
```
totocap, line 163: Do we still have protection from attacks like passing `../../../../file` entries?

packages/mjml-core/src/index.js
```diff
@@ -587,29 +587,14 @@ export default async function mjml2html(mjml, options = {}) {
...
-      }
-      const baseReal = fs.existsSync(base) ? fs.realpathSync(base) : base
       for (const p of pathsArr) {
         if (fs.existsSync(p)) {
           const r = fs.realpathSync(p)
-          const relToBase = path.relative(baseReal, r)
-          const isOutsideBase =
-            !relToBase ||
-            relToBase.startsWith('..') ||
-            path.isAbsolute(relToBase)
```
totocap, line 605: Why did you remove this security?

packages/mjml-cli/src/client.js
DazGreerMJ: Yes, the traversal protection is in `mjml-parser-xml/src/index.js` and was not changed. 

- Null bytes and URL-encoded traversal attempts are stripped by `fullyDecode()`,
- Absolute paths and UNC paths are rejected immediately before any filesystem access,
- The relative path is resolved with `path.resolve(cwd, decoded)`,  so `../../../../file` becomes an absolute path,
- `isPathAllowed(resolvedPath) `then calls `fs.realpathSync` on both the target and each allowed root (following symlinks), and checks the target is inside an allowed root via `path.relative()`. If the result starts with, or is absolute, it's outside and denied.

packages/mjml-core/src/index.js
DazGreerMJ: This was a `console.warn` only, it had no blocking effect. It warned whenever an `includePath` entry resolved outside the template's base directory, which is the intended, normal use case for `includePath` and emitted on every valid use.

The actual path traversal enforcement was not touched and lives entirely in `mjml-parser-xml/src/index.js`. See `isPathAllowed()`.

totocap (approved):

PR #3093, opened 2026-05-21
chore(deps,cli,ci): harden dependency graph, add watch/packaging regressions

dazgreer:
## What's changed

- Dependency updates and resolutions
  - Upgrade CLI watcher stack (`chokidar`, `glob`, `minimatch`)
  - Upgrade `htmlnano` in core
  - Pin `glob` and `cacache` via root resolutions to keep security posture and browser build compatibility
- CLI watch behaviour fix
  - Normalise include dependency paths using realpath resolution so symlinked include edits correctly trigger recompilation
- New regression tests
  - Multi-file watch using `glob-expanded` inputs
  - Symlink include watch recompilation when real target file changes
- CI/release guardrails
  - ~~Add dependency audit job to CI~~
  - Add reusable CLI package smoke test script and package command

## Why

- Remove/mitigate production dependency advisories
- Preserve watch-mode correctness across single-file, multi-file, and symlinked include workflows
- ~~Catch dependency regressions early in CI~~
- Ensure pre-publish CLI artifact is installable and executable in an isolated environment

## Validation performed

- Manual validation previously completed:
  - single-file watch with and without explicit output
  - multi-file and nested glob watch workflows
  - symlink include workflow
- Automated validation:
  - watch regression tests pass (including new cases)
  - full CI passes
  - browser build and smoke test pass
  - ~~dependency audit guard passes~~

## Risk/notes

- Yarn will show resolution mismatch warnings because versions are intentionally pinned above some transitive requested ranges.
- Current validation indicates functional compatibility across CLI, core tests, and browser smoke path.


.github/workflows/mjml-workflow.yml
```diff
@@ -1,6 +1,19 @@
 name: Mjml CI
 on: [push, pull_request]
 jobs:
+  dependency-audit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js 20.x
+        uses: actions/setup-node@v1
```
totocap: Do we want to use more recent options?

.github/workflows/mjml-workflow.yml
```diff
@@ -1,6 +1,19 @@
...
+        with:
+          node-version: 20.x
+      - name: Audit production dependencies
+        run: |
+          yarn install
+          yarn audit:deps
```
totocap: What is the purpose of this? Do we want to make the CI fail when we have security issues in our deps?

package.json
```diff
@@ -5,11 +5,14 @@
     "build:watch": "lerna run build --parallel -- -- -w",
     "build": "lerna run build --parallel --ignore mjml-browser",
     "build-browser": "cd packages/mjml-browser && yarn build",
+    "audit:deps": "yarn audit --groups dependencies",
+    "smoke:cli-pack": "./scripts/smoke-cli-pack.sh",
```
totocap, line 8: Do we want to push this?

.github/workflows/mjml-workflow.yml
dazgreer: The idea was to prevent security regressions from being merged. In hindsight perhaps it's a bit aggressive. I can update this to a warning rather than a fail? That would still allow CI to pass but give us a warning to address. What do you think?

.github/workflows/mjml-workflow.yml
dazgreer: This is dependent on the whether we're keeping the `dependency-audit` job  below. I can update this  if we keep it

Probably should update the `build` and `browser-smoke` jobs regardless though

package.json
dazgreer: the `audit:deps` script is used for the `dependency-audit` job so it depends if we're keeping it. Or I could inline it into the workflow?

`smoke:cli-pack` validates a release path that build, test, and lint don't currently which ispackaging the CLI, installing the tarball in a clean environment, and verifying it actually runs. It's not essential but it's is low-maintenance and useful for catching publish-time regressions. What do you think?

.github/workflows/mjml-workflow.yml
totocap: We can do it in another PR.

.github/workflows/mjml-workflow.yml
totocap: The warnings will not be shown inside the Pull Request view in Github, unless I'm wrong. So I don't think warnings will help. If you want warnings we can add a precommit step

package.json
totocap: Ok

.github/workflows/mjml-workflow.yml
dazgreer: I've removed the dependency-audit job and this script

totocap (approved):

PR #3104, opened 2026-05-28
refactor(mjml-hero): simplify getChildContext, drop dead width parsing

### LINKED ISSUE #2441: This code in mj-hero makes no sense.
I could be really wrong of course but I do not understand this code:

From: https://github.com/mjmlio/mjml/blob/master/packages/mjml-hero/src/index.js#L49

I added my comments to the code.

```
getChildContext() {
    // Refactor -- removePaddingFor(width, ['padding', 'inner-padding'])
    const { containerWidth } = this.context
    const paddingSize =
      this.getShorthandAttrValue('padding', 'left') +
      this.getShorthandAttrValue('padding', 'right')

    // First we format the value to add "px" suffix.
    let currentContainerWidth = `${parseFloat(containerWidth)}px`

    // Now we parse the code again, but why?
    const { unit, parsedWidth } = widthParser(currentContainerWidth, {
      parseFloatToInt: false,
    })

   // We added the "px" suffix a few lines before, how can it ever be "%" unit?
    if (unit === '%') {
      currentContainerWidth = `${
        (parseFloat(containerWidth) * parsedWidth) / 100 - paddingSize
      }px`
    } else {
      currentContainerWidth = `${parsedWidth - paddingSize}px`
    }

    return {
      ...this.context,
      containerWidth: currentContainerWidth,
    }
  }
  ```

### PR DESCRIPTION
**Pull request by danfitz36**
Fixes #2441.

`MjHero.getChildContext` was converting the parent container width to a `px` string and then re-parsing it with `widthParser` to branch on the unit — but the `if (unit === '%')` branch can never fire, because the value is appended with `'px'` on the line above. The whole sequence collapses to a single subtraction.

As @iRyusa noted on the issue, the parsing was a remnant of an earlier attempt to support a `width` attribute on `mj-hero` and is no longer needed.

### Changes

- `packages/mjml-hero/src/index.js`
  - Replace the round-trip parse + dead `if (unit === '%')` branch with a direct `parseFloat(containerWidth) - paddingSize` subtraction
  - Drop the now-unused `widthParser` import
  - Remove the stale `// Refactor -- removePaddingFor(...)` TODO comment
- `packages/mjml/test/hero-padding-inner-width.test.js` (new)
  - Five parametrized cases asserting the inner outlook table width equals `containerWidth − horizontalPadding` for a range of padding shorthands. `mj-hero` previously had no direct test coverage.

### Verification

- Existing suite: 74 → 79 passing (the 5 new cases)
- Output is byte-identical to the old implementation; the dead branch was provably dead
- Lint clean

### DIFF
--- a/packages/mjml/test/hero-padding-inner-width.test.js
+++ b/packages/mjml/test/hero-padding-inner-width.test.js
@@ -1,0 +1,37 @@
+const chai = require('chai')
+const mjml = require('../lib')
+
+describe('mj-hero inner container width', function () {
+  const cases = [
+    { padding: '0', expected: 600 },
+    { padding: '0 20px', expected: 560 },
+    { padding: '0 10px 0 30px', expected: 560 },
+    { padding: '20px', expected: 560 },
+    { padding: '0 50px 0 50px', expected: 500 },
+  ]
+
+  cases.forEach(function ({ padding, expected }) {
+    it(`reduces the inner outlook table width by horizontal padding (padding="${padding}")`, async function () {
+      const input = `
+        <mjml>
+          <mj-body>
+            <mj-hero
+              mode="fixed-height"
+              height="200px"
+              background-width="600px"
+              background-height="200px"
+              padding="${padding}">
+              <mj-text>hero content</mj-text>
+            </mj-hero>
+          </mj-body>
+        </mjml>
+      `
+
+      const { html } = await mjml(input)
+
+      chai
+        .expect(html)
+        .to.include(`style="width:${expected}px;" width="${expected}"`)
+    })
+  })
+})
--- a/packages/mjml/test/hero-padding-inner-width.test.js
+++ b/packages/mjml/test/hero-padding-inner-width.test.js
@@ -1,37 +1,68 @@
 const chai = require('chai')
 const mjml = require('../lib')
 
+const render = (paddingAttrs) =>
+  mjml(`
+    <mjml>
+      <mj-body>
+        <mj-hero
+          mode="fixed-height"
+          height="200px"
+          background-width="600px"
+          background-height="200px"
+          ${paddingAttrs}>
+          <mj-text>hero content</mj-text>
+        </mj-hero>
+      </mj-body>
+    </mjml>
+  `)
+
+// mj-hero emits two Outlook conditional tables: the outer wrapper
+// (align="center" ... role="presentation", always the full container width)
+// and the inner content table whose width is containerWidth − horizontalPadding.
+// Asserting on the inner table specifically avoids false positives where the
+// outer table happens to share the expected width (e.g. zero horizontal padding).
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/mjml/test/hero-padding-inner-width.test.js [hunk 1]
```diff
@@ -0,0 +1,37 @@
...
+            </mj-hero>
+          </mj-body>
+        </mjml>
+      `
+
+      const { html } = await mjml(input)
+
+      chai
+        .expect(html)
+        .to.include(`style="width:${expected}px;" width="${expected}"`)
```
dazgreer: The current test asserts a width substring anywhere in the rendered HTML. Please could you tighten it to target the specific Outlook inner table generated by mj-hero? That would avoid false positives from matching a different table with the same width.

packages/mjml/test/hero-padding-inner-width.test.js
```diff
@@ -0,0 +1,37 @@
...
+    })
+  })
+})
```
dazgreer, line 68: Please consider adding a couple more tests:
- explicit directional padding attributes, since logic prioritises `padding-left`/`padding-right` over the shorthand in shared helper behaviour,
- an edge-case assertion where horizontal padding exceeds container width, to pin the expected behaviour and avoid future ambiguity

dazgreer:
Hi @danfitz36, thanks for your submission! I've reviewed and made a couple of small comments

danfitz36:
did you leave comments? i don't see them

dazgreer (commented):
Thanks for this, just a couple of small comments

packages/mjml/test/hero-padding-inner-width.test.js
danfitz36: Done in 42ecbb7e. The assertion now targets the inner Outlook table specifically — `<table border="0" cellpadding="0" cellspacing="0" style="width:Npx;" width="N" >` — which is distinct from the outer wrapper table (that one carries `align="center"` and `role="presentation"`). This removes the false positive where the outer table shares the expected width when horizontal padding is zero.

packages/mjml/test/hero-padding-inner-width.test.js
danfitz36: Added in 42ecbb7e:
- An explicit `padding-left`/`padding-right` case (no shorthand).
- A case mixing the shorthand with directional attrs, pinning that `padding-left`/`padding-right` take precedence over the shorthand via `getShorthandAttrValue`.
- An edge case where horizontal padding exceeds the container (`padding="0 400px"`), which pins the current negative-width result (-200) so any future clamping is a deliberate, visible change.

guigui64 (approved):

dazgreer (approved):
Great work, thanks!

totocap (approved):

PR #3107, opened 2026-06-05
Add emjay to community ports

julik:
Added information about the emjay Ruby parser and processor for MJML. I needed to experiment with MJML but found the Ruby ports a bit lacking - they all need dependencies, so a quick vibe-port was done based on using Nokogiri to parse MJML as if it were an HTML/XML hybrid. Seems to be working well enough.

### DIFF
--- a/doc/ports.md
+++ b/doc/ports.md
@@ -26,6 +26,12 @@ Native Implemented Function (NIF) bindings for the MJML Rust implementation (mrm
 
 [https://github.com/adoptoposs/mjml_nif](https://github.com/adoptoposs/mjml_nif)
 
+### Ruby: Emjay
+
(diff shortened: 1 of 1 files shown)


dazgreer (commented):
Hi, thanks for the PR for this useful MJML port.

In the README, it says that the `emjay` test suite currently compares its output against the MJML 4 JavaScript implementation. Since we're now on MJML 5, could you please update the tests to use the latest MJML 5 release (5.4.0) and confirm whether they still pass? If so, please also update the README to match.

Also, the URL provided in the update appears to be malformed. Could you please update it?

Thanks

julik:
Thank you for the reply, I'll check all of the above and get back to you.

julik:
Sorry couldn't find the branch I used for this PR to update it - filed a new PR in https://github.com/mjmlio/mjml/pull/3131

PR #3118, opened 2026-07-23
docs: note that templating tags break with mj-html-attributes

Hashim1999164:
## Summary
- Documents that `mj-html-attributes` re-parses HTML and can rewrite template tags that look like HTML/XML (for example EJS `<%= %>`).
- Adds a short clarification on the ending-tags page as well.

Fixes #2744

## Test plan
- [ ] Confirm the new note reads clearly on the mj-html-attributes docs page
- [ ] Confirm ending-tags still makes sense with the extra sentence

Issue #2744, opened 2023-09-13
Issue with mj-html-attributes and templating language tags <% %>


cantoniriccardo:
**Describe the bug**
When we use mj-html-attributes and a templating language like EJS inside a compontent, a tag <%= my_variable %> is not preserved as is, but it is rendered as <%= my_variable="" %="" />.

**To Reproduce**
1. Copy and paste the following code in the try-it-live tool: 
```
<mjml>
  <mj-head>
    <mj-html-attributes>
      <mj-selector path=".custom div">
        <mj-html-attribute name="data-id">42</mj-html-attribute>
      </mj-selector>
    </mj-html-attributes>
  </mj-head>
  <mj-body>
    <mj-section>
      <mj-column>
        <mj-text css-class="custom">
          Hello <%= world %>
        </mj-text>
      </mj-column>
    </mj-section>
  </mj-body>
</mjml>
```
3. See error

**Expected behavior**
We expect the tag <% ... %> to be rendered as is.

**MJML environment**
 - MJML Version 4.9.3
 - MJML try-it-live tool

iRyusa:
Yup, I think this should be mentioned in the doc as it's not something we can fix. When Cheerio parse the document, any "look alike" xml/html will be treated as a xml tag so it will alter the output HTML

PR #3121, opened 2026-07-31
fix(mjml-core): don't turn <br> into <br></br> when an mj-selector is present

### LINKED ISSUE #3112: When any `mj-selector` element is present, `<br>` elements are sometimes compiled to `<br></br>`
**Describe the bug**
`<br>` is getting compiled to different things based on if an `mj-selector` element is present in the MJML. It does not seem to matter if the `mj-selector` actually selects the block the `<br>` is in.

**To Reproduce**
Case 1: Example without `mj-selector`: https://mjml.io/try-it-live/To2DOPgd8

In this example,

```html
...
<mj-text css-class="custom">
  <p>
    Hello World!<br>
  </p>
</mj-text>
...
```

is compiled to

```html
...
<p> Hello World!<br>
</p>
...
```

Case 2: Example with `mj-selector`, `<br>` is compiled to `<br></br>`: https://mjml.io/try-it-live/vjWLQhb0L

In this example,

```html
...
<mj-text css-class="custom">
  <p>
    Hello World!<br>
  </p>
</mj-text>
...
```

is compiled to

```html
...
<p> Hello World!<br>
  </br></p>
...
```

Note that the `path` of the `mj-selector` does not actually select the `mj-text` the `<br>` is in.

Case 3: Example with `mj-selector` but `<br>` is compiled to `<br />`: https://mjml.io/try-it-live/mjLo-DtRrc

In this example, the `<br>` is not placed in a `<p>`.

```html
...
<mj-text css-class="custom">
  Hello World!<br>
</mj-text>
...
```

is compiled to

```html
...
Hello World!<br />
...
```

Again, the `path` of the `mj-selector` does not actually select the `mj-text` the `<br>` is in.

**Expected behavior**
Adding an `mj-selector` should not change what `<br>` is compiled to.

 - MJML Version 5.3.0
 - MJML tool used Try it Live

Ouranos27:
Fixes #3112

### Problem

`<br>` compiles to three different things depending on whether an `mj-selector` exists anywhere in the document, even when its `path` targets something unrelated:

| authored | no `mj-selector` | with an `mj-selector` |
| --- | --- | --- |
| `<br>` inside a `<p>` | `<br>` | `<br></br>` |
| `<br>` as the last child | `<br>` | `<br/>` |
| `<br />` | `<br />` | `<br/>` |

`<br></br>` is not just cosmetically wrong. Clients recover from it inconsistently and some render a doubled line break.

### Cause

`mj-selector` is what populates `globalData.htmlAttributes`, and a non empty `htmlAttributes` round trips the whole document through cheerio in `packages/mjml-core/src/index.js`:

```js
if (!isEmpty(globalData.htmlAttributes)) {
  const $ = load(content, {
    xmlMode: true, // otherwise it may move contents that aren't in any tag
    decodeEntities: false, // won't escape special characters
  })
  // ...apply attributes...
  content = $.root().html()
}
```

XML mode has no notion of HTML void elements, so `<br>` is parsed as an element that requires a closing tag. Whatever follows it inside its parent becomes its children and it is serialized back as `<br>...</br>`; with no children it is serialized as `<br/>`. When no selector is present this block is skipped and the markup is never reparsed, which is why the selector's `path` makes no difference: the reparse rewrites the element, not the selector.

### Approach

`xmlMode: true` is deliberate and the comment on that line explains why, so parsing is left exactly as it is. Only the rendering changes.

Asking cheerio for the HTML rendering does not work, because it picks its renderer from the parsing options. Passing `xmlMode: false` to `$.html()` makes it fall through to parse5, which escapes text and so defeats `decodeEntities: false`, and it does not fix the void elements either since the nodes carry no HTML namespace. Checked locally, `<p>Hi<br>There</p><div>{ if a < 5 }</div>` comes back as `<p>Hi<br>There</br></p><div>{ if a &lt; 5 }</div>`.

`dom-serializer` is the renderer cheerio already uses underneath, and it takes its options directly, so the block renders the document itself instead:

```js
content = render($.root()[0], {
  xmlMode: false,
  decodeEntities: false,
  emptyAttrs: true,
})
```

`xmlMode: false` is what makes the serializer treat void elements as void, `decodeEntities: false` keeps the templating syntax intact for the same reason it is set on the parser, and `emptyAttrs: true` keeps `alt=""` written in full rather than as a bare `alt`.

### Result

Compiled output with a selector present, before and after. This is the whole diff on a template exercising `mj-image`, `mj-divider`, `mj-text`, `mj-social`, `mj-navbar`, `mj-button`, `mj-table`, `mj-accordion`, `mj-carousel` and `mj-spacer`:

```diff
- <p>Hello<br>World!</br></p>
+ <p>Hello<br>World!</p>

- Trailing<br/>
+ Trailing<br>

- <img alt="x" src="..." width="550" height="auto"/>
+ <img alt="x" src="..." width="550" height="auto">

- <input class="mj-accordion-checkbox" type="checkbox" style="display:none;"/>
+ <input class="mj-accordion-checkbox" type="checkbox" style="display:none;">
```

No structural, attribute or entity changes, and conditional comments are untouched.

The first two lines are the bug. The last two are the same change applied to the other void elements, which the serializer treats alike.

Exact parity with the no-selector path is not reachable, because the round trip loses whether the author wrote `<br>` or `<br />`. This block already normalizes that today, writing both as `<br/>`, so what changes here is which form it normalizes to.

### Not covered

The `<%= %>` mangling documented in #3118 / #2744 is a separate, parse level problem (the XML parser reads `<%=` as a tag name). Confirmed locally that this change neither fixes it nor makes it worse.

### Testing

`yarn build && yarn test && yarn lint`. Lint clean. Regression tests added in `packages/mjml/test/html-attributes.test.js`, which compiles each case below with and without the `mj-selector` and asserts the `<br>` output is identical, plus that no closing tag is added to ``. All three fail before this change and pass after.

MJML used to test locally, from the three cases in the issue. Each was compiled twice, once as written and once with the `mj-head` block removed:

**Case 1 and 2, `<br>` inside a `<p>`**

```xml
<mjml>
  <mj-head>
    <mj-html-attributes>
      <mj-selector path=".unrelated div">
        <mj-html-attribute name="data-id">42</mj-html-attribute>
      </mj-selector>
    </mj-html-attributes>
  </mj-head>
  <mj-body>
    <mj-section>
      <mj-column>
        <mj-text css-class="custom">
          <p>
            Hello World!<br>
          </p>
        </mj-text>
      </mj-column>
    </mj-section>
  </mj-body>
</mjml>
```

Before: `<p> Hello World!<br>\n </br></p>`. After: `<p> Hello World!<br>\n </p>`, identical to the no-selector output.

**Case 3, `<br>` outside any `<p>`**

```xml
<mjml>
  <mj-head>
    <mj-html-attributes>
      <mj-selector path=".unrelated div">
        <mj-html-attribute name="data-id">42</mj-html-attribute>
      </mj-selector>
    </mj-html-attributes>
  </mj-head>
  <mj-body>
    <mj-section>
      <mj-column>
        <mj-text css-class="custom">
          Hello World!<br>
        </mj-text>
      </mj-column>
    </mj-section>
  </mj-body>
</mjml>
```

Before: `Hello World!<br/>`. After: `Hello World!<br>`, identical to the no-selector output.

**A `<br>` followed by a sibling element**

```xml
<mjml>
  <mj-head>
    <mj-html-attributes>
      <mj-selector path=".unrelated div">
        <mj-html-attribute name="data-id">42</mj-html-attribute>
      </mj-selector>
    </mj-html-attributes>
  </mj-head>
  <mj-body>
    <mj-section>
      <mj-column>
        <mj-text css-class="custom">
          <p>Hello<br><span>World!</span></p>
        </mj-text>
      </mj-column>
    </mj-section>
  </mj-body>
</mjml>
```

Before: `<p>Hello<br><span>World!</span></br></p>`. After: `<p>Hello<br><span>World!</span></p>`.

I do not have a Litmus or Email on Acid account, so there are no client screenshots here. The change replaces invalid markup with the canonical HTML form rather than adding a styling attribute, so the compiled output above is the evidence. Happy to work with someone who has access if you want client captures before merging.

### DIFF
--- a/packages/mjml-core/package.json
+++ b/packages/mjml-core/package.json
@@ -27,6 +27,7 @@
     "cssnano": "^7.1.2",
     "cssnano-preset-lite": "^4.0.4",
     "detect-node": "^2.0.4",
+    "dom-serializer": "^2.0.0",
     "htmlnano": "^3.3.1",
     "js-beautify": "^1.15.4",
     "juice": "^11.0.0",
--- a/packages/mjml-core/src/index.js
+++ b/packages/mjml-core/src/index.js
@@ -12,6 +12,7 @@ import {
 } from 'lodash'
 import juice from 'juice'
 import { load } from 'cheerio'
+import render from 'dom-serializer'
 import minifier from 'htmlnano'
 import MJMLParser from 'mjml-parser-xml'
 import MJMLValidator, {
@@ -839,7 +840,18 @@ export default async function mjml2html(mjml, options = {}) {
       })
     })
 
-    content = $.root().html()
+    /*
+     * The document has to be parsed as XML, but it is HTML, so it is rendered
+     * back with the HTML serializer. Rendering it as XML would emit void
+     * elements as `<br/>` or, when the parser gave them children, as
+     * `<br></br>`, which means the mere presence of an mj-selector would
+     * change how they are written.
+     */
+    content = render($.root()[0], {
+      xmlMode: false,
+      decodeEntities: false, // same reason as above
+      emptyAttrs: true, // keeps `alt=""` instead of writing it `alt`
+    })
   }
 
   content = skeleton({
--- a/packages/mjml/test/html-attributes.test.js
+++ b/packages/mjml/test/html-attributes.test.js
@@ -107,7 +107,74 @@ describe('html-attributes', function () {
 </mjml>
 `
 
-    const bodies = {
+    const VOID_ELEMENTS =
+      'area|base|br|col|embed|hr|img|input|link|meta|param|source|track|wbr'
+
+    /*
+     * Every void element in the output. Conditional comments are taken out
+     * first, because what is inside them is never parsed and so keeps the form
+     * the component wrote it in.
+     */
+    const voidTags = (html) =>
+      html
+        .replace(/<!--[\s\S]*?-->/g, '')
+        .match(new RegExp(`</?(?:${VOID_ELEMENTS})(?![\\w-])[^>]*>`, 'gi')) ||
+      []
+
+    /*
+     * A template compiled with a selector is parsed and rendered again, one
+     * compiled without a selector is not, so the two outputs are not byte
+     * identical even when they mean the same thing. These two rules absorb the
+     * differences that carry no meaning, so the comparison only fails on a real
+     * one. The form of the tag itself is not left to the comparison, since
+     * `<br/>` reads as `<br>` under the same rule. It is pinned separately, by
+     * the closing tag and self closing assertions below.
+     */
+    const normalize = (tags) =>
+      tags.map((tag) =>
+        tag
+          // `checked=""` and `checked` are the same attribute.
+          .replace(/\s([\w-]+)=""/g, ' $1')
+          // `<img />` and `<img>` are the same element.
+          .replace(/\s*\/>$/, '>'),
+      )
+
+    const expectParity = async (text, name) => {
+      const [without, withSelector] = await Promise.all([
+        mjml(template(false, text)),
+        mjml(template(true, text)),
+      ])
+      const tags = voidTags(withSelector.html)
+
+      // Without this the comparison below would also pass on an output that
+      // lost the element altogether.
+      chai
+        .expect(
+          tags.filter((tag) => tag.startsWith(`<${name}`)).length,
+          `<${name}> is in the output`,
+        )
+        .to.be.above(0)
+
+      chai
+        .expect(
+          tags.filter((tag) => tag.startsWith('</')),
+          `No closing tag on <${name}>`,
+        )
+        .to.deep.equal([])
+
+      chai
+        .expect(
+          tags.filter((tag) => tag.endsWith('/>')),
+          `No self closing slash on <${name}>`,
+        )
+        .to.deep.equal([])
+
+      chai
+        .expect(normalize(tags), 'Same void elements as without a selector')
+        .to.deep.equal(normalize(voidTags(without.html)))
+    }
+
+    const positions = {
       'inside a <p>': `
           <p>
             Hello World!<br>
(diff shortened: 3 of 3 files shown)


dazgreer (commented):
Hi, thanks for this PR! Please could you add a bit more regression coverage around void-element handling?

- Verify parity for multiple void elements with and without `mj-selector`: `br`, `hr`, `input`, `img`.
- Add a conditional-comment case to ensure this stays unchanged: `<!--[if mso]><br><![endif]-->`
- Document assertion normalization in the test so equivalent HTML forms do not cause false failures:
  - `checked vs checked=""`
  - `` vs ``
  
  Thanks

Ouranos27:
Thanks. I have added all three.

The tests now check parity for `br`, `hr`, `input` and `img`. Each element is compiled twice, once with the selector and once without it. The three `<br>` position cases are still there.

The conditional comment has its own test. `<!--[if mso]><br><![endif]-->` comes out byte for byte as it was written, with or without a selector.

The normalization needs a word of explanation, and the test carries it as a comment. The comparison treats `checked=""` as `checked`, and it treats `` as ``. That rule would hide the bug if it stood alone, because it also turns `<br/>` into `<br>`. The form of the tag is therefore pinned before the comparison runs. Two assertions on the output compiled with a selector do that work. The first one requires that no void element carries a closing tag. The second one requires that none carries a self closing slash. The comparison is then left with the attributes and the order.

Conditional comments are stripped before those two assertions run. Their content is never parsed, so it legitimately keeps the slash. The carousel and the accordion both emit `` inside `[if mso]` blocks.

Seven of the eight tests fail on master and pass with the change. The eighth one is the conditional comment test, and it passes either way, because comments were never affected.

dazgreer (approved):
Great... thanks!

dazgreer:
Hi @Ouranos27, thanks for investigating this. However, following some internal discussion we have some concerns with this approach. Namely, we're adding a new dependency (`dom-serializer`) as well as changing the serialisation of all void elements, whereas the actual bug is specific to `<br>`. That means that the HTML output is different depending on whether you include an `<mj-selector>` or not which is not ideal.

I wonder if a better approach might be to protect/replace `<br>` before the Cheerio round-trip and restore it afterwards. We take a similar approach with template tags, e.g. `{{variable}}`, during beautification/minification, which would allow us to scope this to just` <br>`.

Ouranos27:
Thanks for the feedback @dazgreer!

​I'll take a look into the protect/restore approach for <br> tags around the Cheerio parsing step and update the PR once I have a working implementation.

Ouranos27:
Updated the implementation to the protect/restore approach.

`<br>` is swapped for an opaque token before the cheerio round trip and the original tag is written back after `$.root().html()`. Parsing stays in `xmlMode`. `dom-serializer` is gone, and `<hr>`, `<input>` and `` are left to the existing serializer.

The tests now check `<br>` parity for the three positions from the issue, plus authored `<br>`, `<br/>`, `<br />` and `<br clear="all">`. The conditional comment case is unchanged.

One limitation: a selector whose path is `br` cannot add an attribute, because the tag is not in the tree while cheerio runs.

Issue #434, opened 2016-10-31
mj-image missing beginning tbody


robcube:
I had:
```
        <mj-column>
            <mj-image width="450">
              <xsl:attribute name="src">
                <xsl:value-of select="concat($webServerName,'/images/charts/name.png')"/>
              </xsl:attribute>
            </mj-image>
        </mj-column>
```

It renders as:
```
<table role="presentation" cellpadding="0" cellspacing="0" width="100%" border="0">
                    <tbody>
                      <img width="450px">
                        <xsl:attribute name="src">
                             <xsl:value-of select="concat($webServerName,'/images/charts/name.png')"/>
                          </xsl:value-of>
                        </xsl:attribute>
                      </tbody>
                  </table>
```

It is missing tr/td inside the tbody elements, and missing the ending img.

ngarnier:
Hey @robcube, thanks for opening this issue. I'll be honest, I don't really have any knowledge of `xsl` but it seems the issue here is that the `<` & `>` characters are missing with our parser. 

This is a known issue, as detailed here: https://github.com/mjmlio/mjml/issues/427#issuecomment-255693868

Hence, I don't think it's related to `mj-image` but just the use of `xsl` inside `mjml`.

iRyusa:
Hi there, any content inside `mj-image` is not rendered. 

When I test your code, I got : 

```html
                     <td style="word-break:break-word;font-size:0px;padding:10px 25px;" align="center">
                        <table role="presentation" cellpadding="0" cellspacing="0" style="border-collapse:collapse;border-spacing:0px;" align="center" border="0">
                          <tbody>
                            <tr>
                              <td style="width:450px;"><img alt="" title="" height="auto" src="" style="border:none;border-radius:;display:block;outline:none;text-decoration:none;width:100%;height:auto;" width="450"></td>
                            </tr>
                          </tbody>
                        </table>
                      </td>
```

We doesn't support any `xsl` tag as @ngarnier , so I don't think there's a solution for you now

PR #3134, opened 2026-09-10
chore(deps): bump htmlnano to 3.5.0

jrandolf:
Bumps `htmlnano` in `mjml-core` from 3.3.1 to 3.5.0 ([changelog](https://github.com/posthtml/htmlnano/blob/master/CHANGELOG.md)).

## Why

htmlnano 3.3.1 declares a peer dependency of `cssnano ^8.0.0`, while `mjml-core` depends on `cssnano ^7.1.9`. `npm --strict-peer-deps` fails outright with `ERESOLVE`, and pnpm reports `✕ unmet peer cssnano@^8.0.0: found 7.1.9` — a warning by default, an error under `strict-peer-dependencies` — so downstream projects end up carrying a patch to install mjml. htmlnano 3.3.2 widened the range to `^7 || ^8` ([#434](https://github.com/posthtml/htmlnano/issues/434)); 3.5.0 accepts `^7 || ^8 || ^9` and is current. Yarn classic doesn't surface this because it honours `peerDependenciesMeta.optional`, so the mismatch is invisible in this repo's own install.

## This drops Node 20 from CI

htmlnano 3.4.0 moved to `commander` 15 (`engines: node >=22.12.0`) and 3.5.0 to `cosmiconfig` 10 (`engines: node ^22.18 || >=24`). Yarn classic treats an engine mismatch as fatal, so on the current matrix's Node 20 leg:

```
error commander@15.0.0: The engine "node" is incompatible with this module. Expected version ">=22.12.0". Got "20.20.2"
error Found incompatible module.
```

Node 20 reached [end of life on 2026-04-30](https://github.com/nodejs/Release), so this PR drops `20.x` from the test matrix and moves the `browser-smoke` and `npm-publish` jobs to `22.x`. **The publish workflow change is included because it also runs `yarn install`** — without it, releases would break on the next dispatch. Worth a careful look, since it's the release path.

Two things worth knowing:

- The effective floor is **Node 22.18**, set by `cosmiconfig`, not the 22.12 in the error above. CI's `22.x` resolves well past that.
- This is an *install-time* constraint, not a runtime one. I confirmed htmlnano 3.5.0 executes fine on Node 20.20.2 — `commander` is only reachable from htmlnano's CLI bin, never from the library entry point. So npm/pnpm users on Node 20 would keep working; only yarn-classic installs hard-fail.

### Why the matrix stops at 24

Node 26 becomes LTS on 2026-10-28, so it's a natural next addition — but the test suite can't run on it yet, and that's unrelated to this PR. `mocha@10` resolves `yargs@16.2.0`, whose extensionless `node_modules/yargs/yargs` entry Node 26 loads as ESM:

```
ReferenceError: require is not defined in ES module scope, you can use import instead
    at .../node_modules/yargs/yargs:3:69
```

I confirmed this reproduces on unmodified `master` (htmlnano 3.3.1) under Node 26.8.1, so it's a pre-existing tooling issue rather than a regression here. Adding `26.x` would just paint CI red. `install`, `build`, `lint` and `build-browser` all pass on Node 26 — only `test` fails. Worth resolving separately before 26 promotes; happy to open a separate issue or PR for it.

If dropping Node 20 isn't something you want to do in a dependency bump, **3.3.2 is the last version that installs on Node 20** and still fixes the cssnano peer range. Happy to retarget this PR to 3.3.2 instead — just say the word.

## Output behaviour

Since 3.4.0 the `safe` preset enables `minifyCharacterReferences`, which decodes character references to raw UTF-8. mjml emits `&nbsp;` (`mj-divider`) and `&#8202;` (`mj-spacer`), and mjml's own output carries the XHTML namespace on `<html>`. Straight bump, before → after:

```
-  ...line-height:50px">&#8202;</div>            +  ...line-height:50px">⟨U+200A⟩</div>
-  Hello&nbsp;world &mdash; caf&eacute; &amp; co. &lt;tag&gt; &copy;2026
+  Hello⟨U+00A0⟩world ⟨U+2014⟩ caf⟨U+00E9⟩ & co. &lt;tag> ⟨U+00A9⟩2026
```

Raw `U+00A0`/`U+200A` bytes and a bare `&` are exactly what ESP template engines and re-encoding transports mangle, so I've pinned `minifyCharacterReferences: false` in `htmlnanoOptions`. It sits before the `...minifyOptionsRest` spread, so callers can still opt in via `minifyOptions`. Three regression tests cover it; two of them fail without the pin.

With that pin, the only remaining output change across the templates I rendered is one byte per document, from 3.4.0's `minifyAttributes` handling of the viewport meta:

```
- <meta name="viewport" content="width=device-width, initial-scale=1">
+ <meta name="viewport" content="width=device-width,initial-scale=1">
```

Semantically identical and universally parsed, so I left it. `dir="ltr"` stripping (new in 3.5.0) doesn't affect default output — mjml emits `dir="auto"`.

## Verification

Full CI sequence (`yarn install --frozen-lockfile && yarn build && yarn lint && yarn test && yarn build-browser`) green on **Node 22.22.3 and 24.20.0**: 134 tests passing, lint clean (5 pre-existing `no-console` warnings), browser smoke test passes. Node 20 was not tested — it cannot install, which is the point of the matrix change.

Output comparison was done by rendering five templates (divider, spacer, entity-heavy text, comments, and a combined one) with `minify: true` on 3.3.1 and on this branch, both under Node 22, and diffing the bytes. I also rendered the 3.3.1 baseline under Node 20 and 22 to confirm the Node version itself doesn't affect output.

Lockfile changes are scoped to htmlnano and its transitives: `commander` 15, `cosmiconfig` 10, `tinyglobby`, `fdir`, `js-yaml` 5, `picomatch` 4.

### DIFF
--- a/.github/workflows/mjml-workflow.yml
+++ b/.github/workflows/mjml-workflow.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [20.x, 22.x, 24.x]
+        node-version: [22.x, 24.x]
     steps:
       - uses: actions/checkout@v5
       - name: Use Node.js ${{ matrix.node-version }}
@@ -25,10 +25,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v5
-      - name: Use Node.js 20.x
+      - name: Use Node.js 22.x
         uses: actions/setup-node@v4
         with:
-          node-version: 20.x
+          node-version: 22.x
       - name: Build browser bundle + smoke test
         run: |
           yarn install
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -23,7 +23,7 @@ jobs:
           fetch-depth: 0
       - uses: actions/setup-node@v4
         with:
-          node-version: '20.x'
+          node-version: '22.x'
           registry-url: 'https://registry.npmjs.org'
       - run: yarn install
       - run: yarn build
--- a/packages/mjml-core/package.json
+++ b/packages/mjml-core/package.json
@@ -27,7 +27,7 @@
     "cssnano": "^7.1.9",
     "cssnano-preset-lite": "^4.0.6",
     "detect-node": "^2.0.4",
-    "htmlnano": "3.3.1",
+    "htmlnano": "3.5.0",
     "js-beautify": "^2.0.3",
     "juice": "^11.1.1",
     "lodash": "^4.17.21",
--- a/packages/mjml-core/src/index.js
+++ b/packages/mjml-core/src/index.js
@@ -904,6 +904,12 @@ export default async function mjml2html(mjml, options = {}) {
       removeEmptyAttributes: true,
       minifyJs: false,
       removeComments: keepComments ? false : 'safe',
+      // htmlnano's safe preset decodes character references to raw UTF-8
+      // (`&nbsp;` -> U+00A0, `&#8202;` -> U+200A). Email output travels through
+      // ESP template engines and transports that re-encode payloads, where raw
+      // bytes turn into mojibake, so keep the references escaped. Callers can
+      // opt in through `minifyOptions`.
+      minifyCharacterReferences: false,
       ...minifyOptionsRest,
     }
 
--- a/packages/mjml/test/minify-character-references.test.js
+++ b/packages/mjml/test/minify-character-references.test.js
@@ -1,0 +1,68 @@
+const chai = require('chai')
+const mjml = require('../lib')
+
+const NBSP = ' '
+const HAIR_SPACE = ' '
+
+describe('character references when minifying', function () {
+  this.timeout(10000)
+
+  // ---------------------------------------------------------------------------
+  // htmlnano's safe preset decodes character references to raw UTF-8 since 3.4.0.
+  // mjml keeps them escaped so minified output survives ESP template engines and
+  // transports that re-encode the payload.
+  // ---------------------------------------------------------------------------
+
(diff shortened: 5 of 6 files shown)


hubipe:
Adding a data point in favour of merging this: the mismatch is not only an install-time problem for `--strict-peer-deps` / pnpm users. With default npm settings it silently breaks CSS minification at runtime.

**What happens with plain `npm install` (npm 11.8, Node 24, clean directory, `mjml@5.4.1`):**

- npm resolves the conflict by nesting `cssnano@7.1.9` under `node_modules/mjml-core/node_modules/`, so nothing satisfies htmlnano's `cssnano ^8` peer at the top level.
- `htmlnano` then fails its dynamic `import('cssnano')` and prints to stderr:

```
You have to install "cssnano" in order to use htmlnano's "minifyCss" module
```

- The process still exits 0 and the HTML is produced, but with unminified CSS. Anything that treats stderr as an error (we do in our mail pipeline) reports it as a failure.

**Repro:**

```sh
mkdir t && cd t
echo '{"dependencies":{"mjml":"5.4.1"}}' > package.json
npm install --omit=dev
printf '<mjml><mj-body><mj-section><mj-column><mj-text>Hi</mj-text></mj-column></mj-section></mj-body></mjml>' > t.mjml
node_modules/.bin/mjml --config.minify true --config.beautify false t.mjml -s > /dev/null
# -> You have to install "cssnano" in order to use htmlnano's "minifyCss" module
```

Same repro with mjml@5.4.0 (htmlnano ^3.3.1, resolving to 3.4.1) is clean.

Workaround we are using until this lands:

```yaml
"overrides": { "mjml-core": { "htmlnano": "^3.4.1" } }
```

Verified that with the override cssnano 7 is hoisted to the top level and the warning is gone.

PR #3136, opened 2026-09-14
docs(mj-attributes): explain how shorthand and side-specific attributes combine

### LINKED ISSUE #3057: CSS Shorthands in elements do not overrule definitions in <mj-attributes> if they are defined specifically (not as shorthand)
**Describe the bug**
When defining an element's css properties in the <mj-attributes> section, there is a difference in the results, depending on whether the definitions are written as css shorthands or the attributes are defined explicitly.

**To Reproduce**
Compare these two tests:
1  Define section padding via css shorthand: https://mjml.io/try-it-live/5fM2CUrhn
2 Define section padding in four separate attributes: https://mjml.io/try-it-live/jUMYP1j5E

**Expected behavior**
Both examples should behave the same way. The section's definitions should overrule the one in \<mj-attributes>

 - OS: n/a
 - MJML Version 4.xx

lbesecker195:
Closes #3057

As @iRyusa explained in #3057, shorthand attributes and their side-specific versions, such as `padding` and `padding-left`, are separate attributes. The side-specific one always wins, no matter where each is defined. The documentation covers the order in which attributes are applied, but not this rule, and @iRyusa agreed a warning should be added.

This adds a caution box to `packages/mjml-head-attributes/README.md`, right after the existing "Important" box about attribute order. It states the rule, and gives the case from the issue as an example: a `padding-left` set in `mj-attributes` still applies to a section with an inline `padding`.

To make sure the note describes current behavior, I rendered a few sections with the local `master` build and read the generated `td` style:

| `mj-attributes` | inline on `mj-section` | rendered |
|---|---|---|
| `padding-left="40px" padding-right="40px"` | `padding="10px"` | `padding:10px;padding-left:40px;padding-right:40px` |
| `padding="10px"` | `padding-left="40px"` | `padding:10px;padding-left:40px` |
| | `padding-left="40px" padding="10px"` | `padding:10px;padding-left:40px` |
| `mj-all padding-top="30px"` | `padding="10px"` | `padding:10px;padding-top:30px` |
| `border-top="2px solid red"` | `border="1px solid blue"` | `border:1px solid blue;border-top:2px solid red` |

Documentation only, no code changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


packages/mjml-head-attributes/README.md
```diff
@@ -40,4 +40,10 @@ components using `mj-class="<name>"`.
   </ul>
 </div>
 
+<div class="alert alert-caution" role="alert">
```
dazgreer: Please update the wording to the below... 

Shorthand and side-specific attributes

Shorthand attributes such as `padding` or `border` are separate attributes from their side-specific counterparts (e.g. `padding-left` or `border-top``), and are each resolved according to the order above. A side-specific attribute always takes precedence over the shorthand, wherever each one is defined and regardless of the order of attributes within a tag.

For example, setting `<mj-section padding-left="40px;">` and `<mj-section padding="10px">`, either inline or in `mj-attributes`, results in `padding` being set to `10px` and `padding-left` being set to `40px`, with `padding-left` taking precedence.

dazgreer (changes requested):
Hi @lbesecker195, thanks for your contribution. I'm happy to approve with some re-wording, please see my comment.

PR #3139, opened 2026-09-22
fix(mjml-core): anchor integer type matcher to whole value

### LINKED ISSUE #3137: Integer type matcher is unanchored, so cellpadding="6px" (and "abc5") pass validation
## Describe the bug

`cellpadding` and `cellspacing` are declared as `integer`, but the integer type matcher is unanchored `/\d+/`. Any string that *contains* a digit is treated as valid, including `cellpadding="6px"` and `cellpadding="abc5"`.

The value is then copied unchanged onto the HTML `<table>`. `cellpadding` / `cellspacing` are HTML attributes that take a non-negative integer (pixels), not a CSS length. `"6px"` is invalid HTML; some email clients ignore it and you get no padding.

## To Reproduce

```mjml
<mjml>
  <mj-body>
    <mj-section>
      <mj-column>
        <mj-table cellpadding="6px" cellspacing="2px">
          <tr><td>x</td></tr>
        </mj-table>
      </mj-column>
    </mj-section>
  </mj-body>
</mjml>
```

```js
const mjml2html = require('mjml')
const { html, errors } = mjml2html(mjml)
console.log(errors)
// look for cellpadding="6px" in html
```

Also valid today, which should not be:

```mjml
<mj-table cellpadding="abc5">
```

## Expected behavior

A `valid-types` error, matching the documented type (`integer`) and the HTML attribute:

```
Attribute cellpadding has invalid value: 6px for type Integer
```

Bare integers (`cellpadding="6"`) should still pass.

## MJML environment

- MJML Version: current `master` (also 4.x)
- MJML tool used: Node `mjml` CLI / `mjml2html`

## Email sending environment

N/A — this is a validator bug. The rendering risk is that `"6px"` is emitted as-is.

## Affected email clients

N/A for the validator miss. Outlook’s Word engine and other strict parsers can ignore `cellpadding="6px"` entirely.

## Additional context

Root cause is the integer matcher:

```javascript
// packages/mjml-core/src/types/integer.js
this.matchers = [/\d+/]
```

`String.prototype.search` / the type `isValid()` path only needs a digit somewhere in the value.

Suggested fix: anchor it, e.g. `/^\d+$/`. That matches the docs, HTML4/HTML5 (`cellpadding` is a non-negative integer), and what email clients actually honor.

`mj-table` `cellpadding` / `cellspacing` look like the only `integer` attributes, so the blast radius is small. Existing templates that compile cleanly with `"6px"` would start getting a validation error — which is the point.

Yahiro025:
## Summary

Fixes #3137.

The `integer` type matcher was an unanchored `/\d+/`, so any string that merely *contains* a digit (`cellpadding="6px"`, `cellpadding="abc5"`) passed `valid-types` and was copied onto the HTML `<table>`. HTML `cellpadding` / `cellspacing` only accept a non-negative integer.

This change anchors the matcher to `/^\d+$/` so the whole attribute value must be digits. Bare integers like `6` still pass.

## Test plan

- [x] Added `packages/mjml-core/tests/integerType.test.js` covering valid bare integers and the previously accepted invalid forms (`6px`, `abc5`, etc.)
- [x] `yarn workspace mjml-core run test` — 32 passing


pmaldonado-mj:
Hello @Yahiro025

Thank you for your first contribution to MJML! 🙂 I will review this with the team 👍

pmaldonado-mj (approved):

